PR #2, opened 2022-07-13
improve(databend): use latest databend version and correct some numbers

sundy-li:
Thanks for the benchmark, now databend can load the hits data successfully with the latest version(>= [v0.7.125-nightly](https://github.com/datafuselabs/databend/releases/tag/v0.7.125-nightly)).

This pr:

1. Remove outdated scripts, comments, and behaviors. (There are all fixed in the latest version)
2. Correct the wrong `data_size` (because data is loaded atomically. If load is not passed there will be uncommitted blocks on the disk)

We did not test the performance results on fs on `c6a.4xlarge`, we know it works slowly on fs(mainly due to using Async runtime to read low-latency fs  https://togithub.com/tokio-rs/tokio/issues/3664), maybe we will update the results after we solve the issues but it's in low-priority.


alexey-milovidov:
Thank you! I'm going to re-run this and merge.

alexey-milovidov (approved):

databend/benchmark.sh
```diff
@@ -68,38 +68,26 @@ curl https://clickhouse.com/ | sh
 sudo ./clickhouse install
 
 # Load the data
+# Docs: https://databend.rs/doc/learn/analyze-hits-dataset-with-databend
 
-curl 'http://default@localhost:8124/' --data-binary @create.sql
+curl 'http://default@localhost:8125/' --data-binary @create.sql
```
sundy-li, line 73: There is a mistake, `port` should be 8124.

PR #5, opened 2022-07-15
Add support for testing against Singlestore cloud

ruw:
The current results for Singlestore only test a cluster-in-a-box
deployment, which is not a realistic Singlestore deployment. 
This change adds support for running against Singlestore cloud.

I included results for s2 and s24 Singlestore cluster sizes which are
equivalent in core count to the existing c6a.4x and c6a.metal tests.

This also includes a small tweak for the Singlestore run.sh script to
wait for async compilations to complete. This gives the most accurate
'hot' query results as it ensures we have a fully compiled plan.


alexey-milovidov (approved):

alexey-milovidov:
Thank you! s24 is quite impressive!

PR #13, opened 2022-08-01
Update Redshift results using UserID datatype change in schema, and a…

sgromoll:
Adding updated results for Redshift, which includes the following changes:
* Change of UserID column from BIGINT to Numeric(38, 0) data type, which resolves the overflow in query 4.
* The REGEXP_REPLACE query ran without an error in all benchmarks we ran (perhaps this error came from another database, e.g. MonetDB, which has no comment despite failing on this query?).  REGEXP_REPLACE has been supported in Redshift for a long time.
* We are submitting benchmark results for a 192 vCPU 4-node ra3.16large (same vCPU to the ca3.metal instance)


alexey-milovidov (approved):

sgromoll:
Note in case it is also helpful, I modified `run.sh` to also do the create and load to simplify running it (though I didn't include these changes in the PR):

```
#!/bin/bash -eu

TRIES=3

cat create.sql
psql -h ${HOST} -U ${USER} -d dev -p 5439 < create.sql | grep 'Time'

cat load.sql
psql -h ${HOST} -U ${USER} -d dev -p 5439 < load.sql | grep 'Time'

cat queries.sql | while read query; do
    echo "$query";
    for i in $(seq 1 $TRIES); do
        psql -h "${HOST}" -U ${USER} -d dev -p 5439 -t -c 'SET enable_result_cache_for_session = off' -c '\timing' -c "$query" | grep 'Time'
    done;
done;
```

(I also then just added `\timing` to the top of `load.sql` and `create.sql`).

alexey-milovidov (approved):

alexey-milovidov:
Thank you! The HTML will be updated in a few minutes.

PR #21, opened 2022-08-17
Add MongoDB queries and type the inputs correctly

jasonthorsness:
I have ported the queries for MongoDB and also changed the ingest to type the BSON fields more effectively. I checked the query results for correctness against the output of SingleStore for the queries, they seem correct (results are below).

I was not sure how to integrate with the results system - the run.js file currently outputs a JSON with the query number, the iteration number, pass/fail, the time in milliseconds, and the query output (ex: {"q":0,"it":0,"ok":1,"t":319943,"res":[{"c":99997497}]} for the first query).

I am a MongoDB enthusiast but I work for SingleStore.

Query correctness checks:

0  ok
1  ok
2  ok
3  ok
4  ok
5  off by 1 (6019102 vs. 6019103 - NULL counting behavior?)
6  ok
7  ok
8  ok
9  ok
10 ok
11 ok
12 ok
13 ok
14 ok
15 ok
16 ok
17 not comparable (no sort) seems correct
18 ok
19 ok
20 ok
21 ok
22 ok
23 ok
24 seems ok ? ordering not unique enough
25 ok
26 ok
27 ok
28 ok
29 ok
30 ok
31 seems ok ? ordering not unique enough
32 ok
33 ok
34 ok
35 seems ok ? ordering not unique enough
36 ok
37 ok
38 seems ok ? ordering not unique enough
39 seems ok ? ordering not unique enough
40 ok
41 seems ok ? ordering not unique enough
42 ok


alexey-milovidov:
Thank you!

> I was not sure how to integrate with the results system

It is ok to convert the results with some ad-hoc script or even by hand and paste into the result file.

jasonthorsness:
I tried to run this last week on a c6a.4xlarge with 500GiB gp2 but it was really slow (slower than my laptop). I don't think that reflects the performance the database is capable of. Maybe this could be run on a bare metal with fast local SSD, or on Atlas service? I don't have bandwidth to try that at this time, maybe someone else can take a look.

alexey-milovidov:
We can add more than one type of machine to the benchmark.
Yes, you can run on Atlas - this will be an interesting point of comparison.

But baremetal results shouldn't be included because they are difficult to reproduce (difficult to obtain the same machine).
You can use larger VM instead.

jasonthorsness:
For initial results then I'll let it run on m4a.4xlarge with 500 gp2 as that seems to be a standard for a lot of other results - but I imagine better results will come from something with higher IOPS and RAM

jasonthorsness:
It wasn't as bad as I thought it was just getting stuck on query 9 for some reason. I let it complete on AWS c6a.4xlarge 500 GiB gp2 - does this look right?

alexey-milovidov:
Looks alright, thank you!

PR #23, opened 2022-08-19
Add compilation flags for DataFusion

Dandandan:
This builds the release with LTO.

### DIFF
--- a/datafusion/benchmark.sh
+++ b/datafusion/benchmark.sh
@@ -17,7 +17,7 @@ sudo apt install gcc -y
 # Install Datafusion Master Branch
 git clone https://github.com/apache/arrow-datafusion.git
 cd arrow-datafusion/datafusion-cli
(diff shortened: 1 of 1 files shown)


Dandandan:
FYI @waitingkuo

waitingkuo:
@Dandandan thank you! Do you think we should fix the version as datafusion v11 just released. I used master since v10 has some issues to pass the tests

Dandandan:
> @Dandandan thank you! Do you think we should fix the version as datafusion v11 just released? I used master since v10 has some issues to pass the tests

That might be a good one. We might also install just the latest stable version of datafusion-cli.

waitingkuo:
@Dandandan  do you know whether installing from `cargo install datafusion-cli` contains the flag u added?

Dandandan:
> @Dandandan  do you know whether installing from `cargo install datafusion-cli` contains the flag u added?

Yes, those flags are for cargo, it works with building from the crates.io repository as well.

alexey-milovidov (approved):

PR #32, opened 2022-10-04
Add first results from ClickHouse Cloud Beta

alexey-milovidov:
It's open since today, let's run some benchmarks and see what will happen...


alexey-milovidov:
The variability looks related to caches, will investigate later:

Screenshot_20221004_142804

alexey-milovidov:
If we compare using CPU-heavy queries, the sizes line up perfectly:

Screenshot_20221004_143002

PR #43, opened 2022-10-21
ClickHouse results update

nickitat:


clickhouse/create.sql
```diff
@@ -107,4 +107,5 @@ CREATE TABLE hits
     CLID INTEGER NOT NULL,
     PRIMARY KEY (CounterID, EventDate, UserID, EventTime, WatchID)
 )
-ENGINE = MergeTree;
+ENGINE = MergeTree
+SETTINGS index_granularity = 1024;
```
alexey-milovidov, line 111: This should be only for "tuned" results.

nickitat: done

alexey-milovidov (approved):

PR #50, opened 2022-11-14
update results of databend

sundy-li:
Changes of this pr:

1. bench it with the latest `v0.8.111` version, now it can pass all queries.
2. make bench scripts simple
3. update the readme

It's not fast enough but worth to update.


sundy-li:
OMG, I just notice it's tested on gp3, I'll fix this later.

sundy-li:
cc @alexey-milovidov  PTAL

alexey-milovidov (approved):

alexey-milovidov:
Thank you!

alexey-milovidov:
The results are much better than before, like 5 times!

PR #51, opened 2022-11-15
Update DuckDB results to v0.6.0

Mytherin:
This PR updates the DuckDB results to [v0.6.0](https://duckdb.org/2022/11/14/announcing-duckdb-060.html). As we have fixed the issues related to loading the data (duckdb/duckdb#3969), the work-around memory limit modification is no longer necessary.

This also fixes the OOM errors on all of the queries - lowering the memory limit was causing DuckDB to run into the (lowered) memory limit on queries involving large strings, even though the queries complete just fine with the regular memory limit.

We make a number of other modifications:

* Disable preservation of insertion order - this is not required for this benchmark and other systems do not preserve insertion order either
* Enable parallel CSV loading
* Remove the primary key index - DuckDB generally does not need explicit indexes for OLAP workloads
* Replace `LENGTH` with `STRLEN` - in DuckDB `LENGTH` returns the number of unicode codepoints, whereas in ClickHouse it returns the number of bytes in a string. `strlen` is the equivalent DuckDB function for that.
* Call `.fetchall()` in the client to fetch the result - calling `.execute` is not enough to ensure that the query is completed in full and might result in the final pipeline not being completed. 

I have also performed result set verification in a separate branch to ensure that DuckDB and Clickhouse produce the correct results on the queries that have deterministic results. Happy to create a separate PR with the result set verification if desired.


Mytherin:
@alexey-milovidov please have a look when you have time

WoolenWang:
nice job duckdb

alexey-milovidov (approved):

alexey-milovidov:
This is an impressive improvement! Thank you!

PR #53, opened 2022-11-17
Added more versions for ClickHouse versions benchmark

qoega:


alexey-milovidov (approved):

alexey-milovidov:
Cool!

Now we need to add 22.11 as well.

alexey-milovidov:
The cold run is not trustworthy - let's retest everything on a server with gp3 or io volume instead of gp2.

PR #60, opened 2022-12-04
Add duckdb results on c5.4xlarge and c6a.metal

nickitat:


xiaoyong-z:
Hello, @nickitat @alexey-milovidov, thanks for building such a excellent benchmark! 
I'm confused by some of the results on on benchmark board.

Under c6a, metal, and 500gp2, the loading time of clickhouse and duckdb is about 140s. However, the maximum throughput of the gp2 volume is 250MB/s, and the csv/tsv file is about 70GB, which means even if we scan the file at maximum speed, it should take around 280s. 

I'm confused by the results of clickhouse and duckdb, how it loads the file in less than 280 seconds.

image
image

PR #70, opened 2023-01-30
Update s24.json

arno756:
S24 is a cluster size 12 but it's only one instance S24. It looks currently that you need 12 x S24. You just need 1 x S24


arno756:
S24 is a cluster size 12 but it's only one instance S24. It looks currently that you need 12 x S24. You just need 1 x S24

PR #71, opened 2023-01-31
Update s2.json

arno756:
We only use 1 S2 instance and not 2 S2 instances


arno756:
We only use 1 S2 instance and not 2 S2 instances

PR #76, opened 2023-02-14
Update ClickHouse results

nickitat:


clickhouse/results/c6a.metal.json
```diff
@@ -1,58 +1,58 @@
...
+    "date": "2023-02-14",
     "machine": "c6a.metal, 500gb gp2",
     "cluster_size": 1,
     "comment": "",
 
     "tags": ["C++", "column-oriented", "ClickHouse derivative"],
 
-    "load_time": 137.610,
-    "data_size": 14571706777,
+    "load_time": 461.326,
```
alexey-milovidov, line 10: This number is much worse than the previous one.
Let's check if the file was decompressed before the insertion.

nickitat: smth wrong with the ec2 instance. I've rerun the benchmark on another vm

alexey-milovidov:
FYI, gp2 disks have 500 MB/sec bandwidth for a certain period of time, then it is lowered to 166 MB/sec.
We have to start testing when it is still 500 MB/sec, to get reproducible results.

alexey-milovidov (approved):

PR #77, opened 2023-02-15
update datafusion result 

waitingkuo:
it was benchmarked by datafusion v11, updated to v18


nickitat:
jfyi, you need also to update index.html by running `generate-results.sh`.
I will do it by myself this time

datafusion/results/f16s_v2.json
```diff
@@ -11,48 +11,48 @@
...
+[1.397, 0.683, 0.653],
+[0.323, 0.245, 0.273],
+[0.979, 0.725, 0.750],
+[1.003, 0.662, 0.646],
+[2.709, 2.491, 2.502],
+[5.380, 4.981, 4.912],
+[0.358, 0.221, 0.217],
+[0.310, 0.263, 0.246],
+[3.360, 3.095, 2.981],
+[52.658, 52.928, 52.789],
```
nickitat, line 23: are you sure everything was ok with this query? now it is 10x worse.

waitingkuo: thank you @nickitat , will generate index.html next time.

we found the performance drop as well. it might be caused by the new implementation of distinct function. tracked by https://github.com/apache/arrow-datafusion/issues/5313 now.

PR #78, opened 2023-02-16
Update databend's result

sundy-li:
The main changes in this PR include:

1. Update the release version of databend to `v0.9.39-nightly`. Compared with the previous version, it has improved a lot.
2. Fix incorrect SQL script, `REGEXP_REPLACE` will use `$1` instead of `\1` to indicate the match group in databend.

3. Add c6a.metal results


sundy-li:
cc @nickitat  PTAL if you are available, thanks!

databend/create.sql
```diff
@@ -106,4 +106,4 @@ CREATE TRANSIENT TABLE hits
     URLHash BIGINT NOT NULL,
     CLID INTEGER NOT NULL
 )
-CLUSTER BY (CounterID, EventDate, UserID, EventTime, WatchID);
+CLUSTER BY (CounterID, EventDate, UserID, EventTime, WatchID) COMPRESSION = "lz4" STORAGE_FORMAT = "native";
```
nickitat, line 109: is it actually correct to consider this configuration default? tuned configurations should be separated from "vanilla" set ups

sundy-li: Because this test is under `storage.type = fs`, but databend is mainly optimized for cloud object storage (default zstd compression, parquet format storage), also the `config.toml` basically does not have too many additional settings.

Compared to other db `DISTRIBUTED BY HASH(UserID ) BUCKETS 192`, these configurations are more like specifying partition keys and engines.

It's easy to use adaptive default configurations by `storage.type`, maybe in the next version a few weeks later.

I don't want this to be a tuning result, so I set this pr to draft.
If you don't think this is a tuning result now, you can merge it, or wait for our next update. Thank you.

sundy-li:
continue in #82

PR #80, opened 2023-02-19
Mariadb improvements

patricklauer:
* Increase innodb buffer pool size
* Insert in small batches to avoid undo log slowdowns

This should improve IO performance substantially.


nickitat:
tuned configuration should be uploaded separately from default one

nickitat:
pls also upload improved benchmark results

alexey-milovidov:
I think it's ok to count this configuration as normal instead of "tuned", because MySQL is known for inappropriate defaults, and the docs suggest tuning the innodb buffer pool size. In other words, the innodb buffer pool size is what every DBA will change.

PR #82, opened 2023-02-24
Update databend's result to v0.9.53 version

sundy-li:
The main changes in this PR include:

1. Remove all tuning parameters, always run the benchmark by default configuration setup.
2. Update the release version of databend to `v0.9.53-nightly`.  A huge improvement over the previous version.
3. Fix incorrect SQL script, `REGEXP_REPLACE` will use `$1` instead of `\1` to indicate the match group in databend.
4. Add c6a.metal and c5.4xlarge results.


sundy-li:
cc @nickitat PTAL if you are available, best wishes!

sundy-li:
Temporarily making it to be draft.

Will add c5.4xlarge results and fix the performance degression of TOPK queries in c6a.metal(mostly Q23, Q24) today.

sundy-li:
Databend did not support `partition by` for convenience, so `Q17` and `Q10` work much slower than other DBMS which use `UserID` as the partition key.  Especially for Q17, if it is partitioned by `UserID`, you can optimize the io processor to read only one partition.

But the overall performance looks good now.

nickitat (approved):
am I right that improvements are mostly provided by the recent [cache enchantments](https://github.com/datafuselabs/databend/pull/9772)?

sundy-li:
@nickitat  Thanks for your time to verify.

> am right that improvements are mostly provided by the cache enhancements, such as https://github.com/datafuselabs/databend/pull/9772?

I'm afraid it's incorrect, actually [table data cache](https://github.com/datafuselabs/databend/pull/9772) is the original parquet data cache, we did not enable this by default. It's used to cache object storage into local disk and memory because remote io is heavy than local.

There is no need to cache data in clickbench since most data will be in page cache during hot query.  
You can use SQL to check if the cache is enabled (currently only meta cache is enabled). 

```
MySQL [(none)]> select name, value, default from system.settings where name  like 'enable_query_result_cache';
+---------------------------+-------+---------+
| name                      | value | default |
+---------------------------+-------+---------+
| enable_query_result_cache | 0     | 0       |
+---------------------------+-------+---------+

MySQL [(none)]> select * from system.configs where name like '%cache%';
+-------+----------------------------------------+-------+-------------+
| group | name                                   | value | description |
+-------+----------------------------------------+-------+-------------+
| cache | enable_table_meta_cache                | true  |             |
| cache | enable_table_bloom_index_cache         | true  |             |
| cache | data_cache_storage                     | none  |             |
| cache | table_data_cache_population_queue_size | 65536 |             |
+-------+----------------------------------------+-------+-------------+
4 rows in set (0.047 sec)
```

But we have enabled `enable_table_meta_cache`, it's a must to have in every database and enabled since the very old version.
 
`table_meta_segment_count ` (similar to `MergeTreeMetadataCache` and `PartMetadata` in clickhouse). The data contains :

1. Data locations
2. Column metas (min/max indexes, bloom_filter_index_location)
3. Summary Statistics

It's a JSON format and not compressed, so it's very large (11MB), we really need to improve it into binary or other compressed formats. [Example file](https://gist.github.com/sundy-li/56c839b22e7ce0f6f0b7882b8b5ef0b5)


If `table_meta_segment_count` is disabled, we need to fetch `11MB` metadata from disk multi times in one query (planner, Segment prunner, Block Pruner, Page Pruner ), so it'll be very slow.  `table_prune_partitions_count` cached the prune part results can also help this.

We also enabled `enable_table_bloom_index_cache` to prune queries with eq condition (this is the reason why some cold query runs slower than others because it must load large bloom indexes).

sundy-li:
The biggest reason for performance improvement is that we have introduced a [new expression framework](https://togithub.com/datafuselabs/databend/pull/9411) in v0.9, so that all operators are type static after `type_checker`, and all functions have [`Domain` attributes](https://github.com/datafuselabs/databend/blob/main/src/query/functions/src/scalars/datetime.rs#L963-L981
). 


Based on this, we can reuse this expression to do [constant folding](https://github.com/datafuselabs/databend/blob/main/src/query/storages/common/index/src/range_index.rs#L80-L107) to filter data (segment pruner, block pruner, page pruner)

Many improvements were found through `clickbench`, I will list them in another PR.

Issue #92, opened 2023-03-23
Inaccurate table size calculation of Mysql


hoptical:
At the moment, [the below command](https://github.com/ClickHouse/ClickBench/blob/732c3cff405f02fcda09f908db8579ce83232939/mysql/benchmark.sh#L23) is used in order to calculate the storage size of the table in MySQL:
```bash
sudo du -bcs /var/lib/mysql
```
Because of the irrelevant files of the Mysql (about 1G) and Binlog files (tens of Gigabytes), it doesn't seem reasonable to me to calculate table size in the mentioned way. Instead, it may be more appropriate to use the below query:
```sql
SELECT 
    table_name AS `Table`, 
    round(((data_length + index_length) / 1024 / 1024), 2) `Size in MB` 
FROM information_schema.TABLES 
WHERE table_schema = "test"
    AND table_name = "hits";
```
This will give us the actual size of the hits table. For 10M rows of the table, I got 14G for the first command and 6.4G for the proposed query. I found out that about 6G of 14G was due to the Binlog files, which are ephemeral, and I don't see any point to include them in the storage result.

Consequently, for the full hits table, we may end up with at most 70G instead of 160G, which is more rational to me.

What do you think?

alexey-milovidov:
Yes, that will be better; let's fix it.

alexey-milovidov:
@hoptical, I've updated the results. Now the table sizes are correct.

PR #93, opened 2023-03-30
Mysql table size calculation

hoptical:
Regarding issue #92, I modified the table size calculation code in mysql-myisam and mysql folders. 
I can't run the benchmark for the whole dataset hence I can't modify the final result number.

Please check and inform me in case of any needed changes.


alexey-milovidov:
Looks good to me. It should also be applied for MariaDB and Infobright.

hoptical:
> Looks good to me. It should also be applied for MariaDB and Infobright.

@alexey-milovidov Applied on MariaDB and Inforbright.

alexey-milovidov:
Ok, now we have to re-run the benchmarks...

hoptical:
@alexey-milovidov Unfortunately, I can't run the benchmarks.

alexey-milovidov:
No problem, we can do it.

PR #123, opened 2023-07-11
Tune the Greenplum

adam8157:
"DISTRIBUTED RANDOMLY" is not recommended, which is supposed to give the
worst results.


adam8157:
Hi, could someone run it and update the results? I didn't because I only have GCP but no AWS access, sorry.

qoega:
>which is supposed to give the worst results.

Why? ClickBench reports first query as cold and fastest as hot. It should not matter if you use the same node second time at least once.

adam8157:
> > which is supposed to give the worst results.
> 
> Why? ClickBench reports first query as cold and fastest as hot. It should not matter if you use the same node second time at least once.

It's not about hot or code.

Greenplum's model is share-nothing, every node stores its dedicated data, if the data were distributed randomly among the nodes, the performance would not be good because it could not push the computing to segments without cost.

Instead, this PR changes the distribution to the hash, the planner could take advantage if the query fits, push the computing to segments and get a much better performance result.

PR #141, opened 2023-09-27
Update DuckDB to v0.9 and remove several outdated settings and comments

Mytherin:


nickitat:
Hi @Mytherin, thanks for the update
could you pls also update `index.html`, for this you need to run `generate-results.sh` from the root of CB repo

nickitat (approved):

PR #142, opened 2023-10-26
Update QuestDB result, fixed schema, queries

ideoma:
- Updated schema to use `short` instead of `int` for numeric types where other DBs use `SMALLINT`. Changed `string` to `symbol` for categorical types.
-  Changed filtering to use `NOT NULL` instead of `<> ''` to get same results as other DBs because QuestDB types are nullable and empty values in CSV are imported as `NULL`
- Changed `LIMIT 1000,10` to be `LIMIT 1000, 1010` which is the correct equivalent of the queries used for other DBs
- Updated to latest QuestDB version 7.3.3 on `c6a.metal` aws Ubuntu 22.04 box


nickitat:
could you pls also update `index.html`, for this you need to run `generate-results.sh` from the root of CB repo

ideoma:
Need some cleanup, will resubmit

PR #145, opened 2023-11-19
DataFusion v33 update and add benchmark for partitioned parquet

kmitchener:
Updated for DataFusion v33


index.html
nickitat, line 1: diff here is unjustifiably large, I cannot verify that everything is ok.
could you pls fix it.

datafusion/results/single.json
```diff
@@ -1,58 +1,58 @@
...
+[1.529, 0.117, 0.113],
+[2.265, 0.832, 0.836],
+[2.652, 1.246, 1.247],
+[0.091, 0.065, 0.071],
+[0.100, 0.063, 0.063],
+[2.465, 1.490, 1.478],
+[3.672, 2.539, 2.496],
+[1.627, 0.344, 0.330],
+[1.885, 0.366, 0.369],
+[2.679, 1.253, 1.236],
```
Dandandan, line 26: Quite some queries regressed, are you sure you are using the same setup as in 
[jychen7](https://github.com/jychen7)
(c6a.4xlarge, 500gb gp2) ?

kmitchener: I think it was gp3 storage actually, which is what new instances default to. let me re-run this again and confirm the gp2 500gb storage. I'll put this in draft until I update this and for nickitat's comment.

index.html
kmitchener: fixed! I used the script but extracted just the update for DataFusion for index.html.

PR #147, opened 2023-11-24
Update QuestDB to 7.3.5 and improve schema

puzpuzpuz:
* Updated schema to use `short` instead of `int` for numeric types where other DBs use `SMALLINT`. Changed `string` to `symbol` for categorical types.
* Changed filtering to use `NOT NULL` instead of `<> ''` to get the same results as other DBs because QuestDB types are nullable and empty values in CSV are imported as `NULL`.
* Changed `LIMIT 1000,10` to `LIMIT 1000, 1010` which is the correct equivalent of the queries used for other DBs.
* Removed compilation time from QuestDB result collection scripts since execution time includes compilation.
* Updated results to the latest QuestDB version 7.3.5.


puzpuzpuz:
@nickitat could you take a look at this one? We closed #142 since we wanted to do a QuestDB release first and then update the results. This PR should be ready for review and we made sure to run `./generate.sh` to update the index page.

questdb-partitioned/results/c6a.metal.json
```diff
@@ -1,55 +1,55 @@
...
+        [3.00846,2.92276,2.92005],
+        [9.27325,9.24496,9.3065],
+        [12.9493,12.6288,12.6471],
+        [12.4508,12.4553,12.2962],
+        [20.2326,19.9919,20.2253],
+        [0.0580621,0.0343657,0.0371066],
+        [71.8848,0.327761,0.276862],
+        [0.291733,0.278372,0.296456],
+        [47.8758,0.251057,0.267674],
+        [13.9825,0.020364,0.104363],
```
nickitat, line 34: I'm a little bit sceptical that Q23 could be executed in 20ms without query result caching or smth similar. is everything ok with this measurement?

questdb/run.sh
```diff
@@ -13,7 +13,7 @@ cat queries.sql | while read query; do
     echo "$query";
     for i in $(seq 1 $TRIES); do
         curl -sS --max-time 600 -G --data-urlencode "query=${query}" 'http://localhost:9000/exec?timings=true' 2>&1 | grep '"timings"' ||
-            (questdb-7.3-rt-linux-amd64/bin/questdb.sh stop >/dev/null 2>&1; questdb-7.3-rt-linux-amd64/bin/questdb.sh start >/dev/null 2>&1; sleep 5; echo 'null')
+            (questdb/bin/questdb.sh stop >/dev/null 2>&1; questdb/bin/questdb.sh start >/dev/null 2>&1; sleep 5; echo 'null')
```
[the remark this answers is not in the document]
puzpuzpuz: Initial version of these scripts was contributed by Alexey Milovidov and these restarts were already there. I was under the impression that all databases are put in the same conditions and there are restarts in other scripts too (didn't check them). Isn't that the case?

questdb-partitioned/results/c6a.metal.json
puzpuzpuz: We don't cache anything except for the query execution plan, so, considering server restarts between query executions, the only cache used in hot query executions is the OS page cache. Let me explain why it takes 20ms to execute this query in QuestDB.

Q23 is this query:
```sql
SELECT * FROM hits WHERE URL LIKE '%google%' ORDER BY EventTime LIMIT 10;
```

QuestDB is a time-series database, so table data is sorted by a [timestamp column](https://questdb.io/docs/concept/designated-timestamp/) (`EventTime` in this case). Thus, the `ORDER BY EventTime` clause is basically a no-op. To execute this query we simply filter the table on [multiple threads](https://questdb.io/blog/2022/05/26/query-benchmark-questdb-versus-clickhouse-timescale/) and finish the query as soon as we find 10 rows.

nickitat: > QuestDB is a time-series database, so table data is sorted by a [timestamp column](https://questdb.io/docs/concept/designated-timestamp/)

ok, wasn't expecting that

questdb/run.sh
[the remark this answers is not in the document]
nickitat: > there are restarts in other scripts too (didn't check them). Isn't that the case?

no, that's why it caught my eye. prob he had to do it, because many queries ran with some errors.

[compatible/questdb/result/c6a.4xlarge.txt](https://github.com/ClickHouse/ClickBench/commit/bdd944b584f030731814b6c67156ed798525a59f#diff-12995188669d484ec2bf2ce1ecc42aac29bd678f4da04412b1dfe3995f00a01f)

if it is not the case now, I think it is better to remove restarts

questdb-partitioned/results/c6a.metal.json
nickitat: then I'm wondering if everything is ok with 4xlarge measurement for Q23

> [151.752,153.277,153.319]

questdb/run.sh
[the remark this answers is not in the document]
puzpuzpuz: Oh, good to know the reason.

> of course we don't have to do it in this pr.

We'll get rid of these restarts in the next ClickBench result update.

questdb-partitioned/results/c6a.metal.json
puzpuzpuz: > then I'm wondering if everything is ok with 4xlarge measurement for Q23

Yes, that result is expected to be slow since the table there is non-partitioned (it doesn't have designated timestamp), i.e. there is no order in the on-disk data. Hence, the query engine has to do a full filter scan to process this query there.

puzpuzpuz:
@nickitat thanks for the review!

nickitat:
sure thing, was interesting to learn more about your system

puzpuzpuz:
> QuestDB was surprisingly unstable when I added it to ClickBench: memory leaks, segfaults, infinite loops, and wrong results. That's why I've added a server restart after every query. It's probably not relevant anymore.

Yes, we indeed had a few bugs in old QuestDB versions preventing ClickBench from a successful run. Thanks for the confirmation: we'll remove the redundant restarts in the next result update.

PR #160, opened 2024-01-16
Update QuestDB result to v7.3.9

puzpuzpuz:
Also fixes LIMIT values in one of queries and removes server restarts between query executions.


puzpuzpuz:
Hi @nickitat

Could you take a look at this one? It includes server restart removal we [discussed](testChunkedDataIlpUploadNoKeepAlive) previously.

puzpuzpuz:
@nickitat many thanks for the snappy review!

PR #181, opened 2024-04-09
Updated Oxla results

adam-szymanski:
I have run newest version of Oxla (1.20) and have created an index that exactly matches ordering of table definition in Clickhouse.


alexey-milovidov:
@tinybit PTAL.

melvynator (approved):
@adam-szymanski Thank you for your contribution. 

I was able to reproduced your test. 

Means Comparison:
Your submission 1:
Column 1: 1.929
Column 2: 1.244
Column 3: 1.249
My replication:
Column 1: 2.072
Column 2: 1.676
Column 3: 1.363

The script worked well, the only hiccup I faced was around the download of the data that for some reason I was not able to attached to the docker container probably and I had to download it directly from the container. 

But I think we should be good to merge.

PR #190, opened 2024-05-16
Add initial benchmarking for Opteryx

joocer:
> Thank you for the contribution!
> 
> Please make sure that you've updated `index.html`. In order to do this run `generate-results.sh` from the root directory of ClickBench repository.

I've updated the index.html by running the script, but it doesn't render correctly - I'm not sure if that's okay as it looks like the last update to the index was done by a bot, so maybe it does something to make it render okay?

Thank you.

Joocer


auxten (changes requested):
The index.html is broken in your PR. You should reset it and run the `./generate-results.sh` again.

joocer:
Thanks @auxten , I'll address these issues in a new PR.

PR #193, opened 2024-05-29
QuestDB 8.0

puzpuzpuz:
Replaces STRING column type with the new VARCHAR type. Also, changes the type of a few boolean flag columns from SHORT to BYTE.


puzpuzpuz:
Hi @nickitat 

Could you take a look at this PR?

questdb-partitioned/create.sql
```diff
@@ -1,73 +1,73 @@
 CREATE TABLE hits
 (
     WatchID long,
-    JavaEnable short,
-    Title string,
-    GoodEvent short,
+    JavaEnable byte,
```
nickitat, line 4: sorry for an obvious question, but is `byte` big enough to represent all the possible values?

puzpuzpuz: Yes, this and many other columns are boolean-like, i.e. they take only 0 and 1 as their values. I've checked distinct value for all columns before changing their type.

puzpuzpuz:
@nickitat many thanks for the review!

PR #195, opened 2024-06-03
Add Presto

Archy-X:
This PR adds results for Presto. https://prestodb.io/


presto/create.sql
```diff
@@ -0,0 +1,107 @@
...
+    UTMMedium VARCHAR(65535) NOT NULL,
+    UTMCampaign VARCHAR(65535) NOT NULL,
+    UTMContent VARCHAR(65535) NOT NULL,
+    UTMTerm VARCHAR(65535) NOT NULL,
+    FromTag VARCHAR(65535) NOT NULL,
+    HasGCLID SMALLINT NOT NULL,
+    RefererHash BIGINT NOT NULL,
+    URLHash BIGINT NOT NULL,
+    CLID INT NOT NULL
+);
```
yingsu00, line 107: Would it be easier for the users if we add the WITH clause here?

presto/create.sql
```diff
@@ -0,0 +1,107 @@
+CREATE TABLE IF NOT EXISTS hits (
+    CounterID INT NOT NULL,
```
yingsu00, line 2: The schema does NOT match the Thrift message schema:
```
File path:  Downloads/hits.parquet
Created by: parquet-cpp version 1.5.1-SNAPSHOT
Properties: (none)
Schema:
message schema {
  required int64 WatchID;
  required int32 JavaEnable (INTEGER(16,true));
  required binary Title (STRING);
  required int32 GoodEvent (INTEGER(16,true));
  required int64 EventTime;
  required int32 EventDate (INTEGER(16,false));
  required int32 CounterID;
  required int32 ClientIP;
  required int32 RegionID;
  required int64 UserID;
...
```
Can you please double check the schema and fix the DDL here?

presto/create.sql
allenshen13: Fixed now.

presto/create.sql
allenshen13: Fixed now.

presto/README.md
```diff
@@ -0,0 +1,28 @@
+# Presto
+
+Presto is a distributed SQL query engine for big data.
+- [Github](https://github.com/prestodb/presto)
+- [Homepage](https://prestodb.io)
+
+The benchmarks are based on Presto version `0.287`.
+
+We assume that a Presto cluster is already running. For more information, visit [Getting Started](https://prestodb.io/getting-started/).
```
alexey-milovidov, line 9: Need to add instructions on how to install Presto on an empty machine with Ubuntu on AWS.

presto/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+----------
+## Steps
+
+1. Download the parquet file and upload it to an S3 Bucket ex. s3://your-bucket/clickbench-parquet/hits/hits.parquet.
```
alexey-milovidov, line 14: We already have it,

`https://clickhouse-public-datasets.s3.eu-central-1.amazonaws.com/hits_compatible/athena_partitioned/hits_{0..99}.parquet')`

allenshen13: Okay, made a note of this in the README.

presto/README.md
allenshen13: It's probably not possible to have a single script to set up a cluster since the environment is complicated, what should I do?

chhetripradeep: Having a standard setup is important otherwise different setups may not give identical results.

Would be nice if you can do either of the following:
- You can create an AWS AMI and add that AMI id in the README. 
- You can setup a docker-compose based setup of Presto.

rschu1ze: @allenshen13 As the other commenters said already, `benchmark.sh` is supposed to be as automatic as possible. This ensures reproducibility and comparability with other systems.

Looking at https://prestodb.io/docs/current/installation/deploy-docker.html, it seem fairly easy to deploy a single-node Presto cluster. The JVM config file could be stored within /presto and copied to the right place by "benchmarks.h". It would be great to have Presto included, would you like to continue this PR?

rschu1ze:
@Archy-X Still interested in completing this PR? Otherwise, please close it, thanks.

alexey-milovidov:
@rschu1ze, we do want to have Presto and Trino, as well as Starburst Cloud, in ClickBench.

Archy-X:
I am no longer able to work on this, but @wanglinsong or @xpengahana may be able to provide an update/take over.

alexey-milovidov:
I'm trying to add a similar system, Trino.
No matter what I'm trying, it always prints `Could not read table schema`.
Documentation does not help, and AI does not help either.

alexey-milovidov:
Added: https://benchmark.clickhouse.com/#system=+curp|H(u,s|rsoP%20t|etrs|Ttt|Tts&type=-&machine=-6t|ca2l|g4e|6ax|6ale|3al&cluster_size=-&opensource=-&hardware=+c&tuned=+n&metric=combined&queries=-

PR #196, opened 2024-06-06
Add Crunchy Bridge for Analytics

onderkalaci:
Adding results for Crunchy Bridge for Analytics (a managed service):
   https://www.crunchydata.com/products/crunchy-bridge-for-analytics


crunchy-bridge-for-analytics/create.sql
```diff
@@ -0,0 +1,118 @@
+-- Auto-infer the columns from the hits.parquet
+CREATE FOREIGN TABLE hits_auto () SERVER crunchy_lake_analytics OPTIONS (path 's3://clickhouse-public-datasets/hits_compatible/hits.parquet');
+
+-- To show representative behavior of Crunchy Bridge for Analytics, we point to the
+-- official hits.parquet file on S3 and load it to a local disk before running the queries
+SELECT * FROM crunchy_file_cache.add('s3://clickhouse-public-datasets/hits_compatible/hits.parquet');
```
alexey-milovidov, line 6: Ok, but need to split cold and hot results.
Cold results should not use the cache.

marcoslot: Thanks for chiming in. We can change it that way if preferred, but the cache being referred to here is fully loading the data to disk rather than something like a buffer manager. 

Hence we included load time and felt it would be more accurate to have separate modes (query directly from S3 - no load time, query from local disk - with load time), where this PR adds the results of the latter.

onderkalaci: Hi @alexey-milovidov,

Have you had a chance to review Marco's comment above?

To clarify, the term `cache` here means that we download the files from S3 to the local machine's disk running the benchmark. That's why we use the title `"system": "Crunchy Bridge for Analytics (Parquet, local)"` to clearly describe our process.

In the future, we may add `"system": "Crunchy Bridge for Analytics (Parquet, remote)"`, where we query the Parquet file directly on S3.

Regarding your point:

> Cold results should not use the cache.

If you still believe we should avoid loading the data to the disk on the first run, we might need to adopt a more complex approach:

1. Disable auto-loading of the files to the disk (non-default behavior)
2. Execute the first run.
3. Load the files to the disk.
4. Execute the second and third runs.

Looking forward to your thoughts.

chhetripradeep: Hi @onderkalaci 

As you mentioned, it is just a downloaded copy of s3 file on local disk, will an update query update both the local disk file as well as remote s3 file ?

chhetripradeep: Reading through the Crunchy Bridge docs, I see two sections:

- Data Import & Export: https://docs.crunchybridge.com/analytics/data-import-and-export
- Caching: https://docs.crunchybridge.com/analytics/caching

May i ask what is the difference between the two.

onderkalaci: Hi @chhetripradeep 

> Hi @onderkalaci
> As you mentioned, it is just a downloaded copy of s3 file on local disk, will an update query update both the local > disk file as well as remote s3 file ?

As of the current release of Crunchy Bridge for Analytics, the system is read-only , you specify a file on `S3` and then query it.

> Reading through the Crunchy Bridge docs, I see two sections:
Data Import & Export: https://docs.crunchybridge.com/analytics/data-import-and-export


Crunchy Bridge for Analytics is an extension to Postgres. It means that in the same database cluster, you can have both regular Postgres tables (e.g., heap tables) as well as Crunchy Bridge for Analytics tables, the latter is provided via the extensions.

Just to be clear, in this benchmark, we are only running the tests against the Crunchy Bridge for Analytics tables.

Data Import piece is a capability for regular Postgres tables, not for analytics tables. It gives the ability to pull data into regular Postgres tables from files in s3 in formats like `parquet`, `json` or `csv`.


Data Export is the ability to push data from any Postgres data source (including regular Postgres tables as well as Crunchy Bridge for Analytics tables, or any query results etc) to `S3` in formats like `parquet`, `json` or `csv`.

So using data import feature and then running ClickBench would correspond to the vanilla postgresql benchmarks in https://github.com/ClickHouse/ClickBench/tree/main/postgresql , using the regular PostgreSQL storage and query engine.

> Caching: https://docs.crunchybridge.com/analytics/caching

Caching is for Crunchy Bridge for Analytics tables. As we described in the previous comments, `Crunchy Bridge for Analytics` tables are used to query files in `S3`. In many real use-cases, it is convenient to cache the data (e.g., download the files to the local machine) to prevent querying over the network repeatedly. This also happens automatically, but asynchronously in the background. Ongoing queries use (remote) range requests to `S3` until the data is available locally, so it is not like a buffer cache where the data is cached after the first query. It is effectively two modes of operation.


And, for this test, we use `"system": "Crunchy Bridge for Analytics (Parquet, local)"`  as the title such that we download files upfront and make them local. This is already the experience that many users have with the system.


Let us know if anything is missing.

chhetripradeep: Thank you for the response @onderkalaci

As per [ClickBench Rules](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#rules-and-contribution), the first run of a query needs to be cold-run and second & third runs of the query can be hot-run, we would recommend to drop cache after every third run of a query. We need to ensure that we are following same rules for all databases otherwise results gets biased. So my understanding of the order of events is:

For each query:
- Execute the first run of the query. I guess It will auto-load the cache.
- Execute the second & third run of the query.
- Drop cache using `select crunchy_file_cache.remove('..')`

chhetripradeep: Also i ran the scripts and got this [result](https://pastila.nl/?01782bfe/6b4ad825e86c30da0bfb02e1b2a34b3a#BaxmFv9DaOvsv5hiwTgBUg==) The result is bit different from the one from the PR i guess because of `psql timing` measure which is measuring client side latency. Would be good if you can add the location of client from which you expect to run the scripts from.

onderkalaci: hi @chhetripradeep,

Alright, let me re-run the tests, though the following is not 100% accurate.  The file download happens asynchronously, so without explicitly loading the file, the second run could still be "cold"

> For each query:
Execute the first run of the query. I guess It will auto-load the cache.
Execute the second & third run of the query.
Drop cache using select crunchy_file_cache.remove('..')

Also, "loading" / "dropping" the local file per query would make the running of the test quite long. So, instead we'd have something like this:
- Disable cache
- Run the first execution of all queries
- Enable cache & Synchronously load the data
- Run the second and the third executions of the queries

Are you OK with that?

Thanks!

chhetripradeep: > Disable cache & run the first execution of all queries

If these steps make sure that first execution of all queries doesn't pick any data from cache, we are good to go with this approach.

onderkalaci: > > Disable cache & run the first execution of all queries
> 
> If these steps make sure that first execution of all queries doesn't pick any data from cache, we are good to go with this approach.

Yes, this makes sure the first execution of all queries use remote file.

onderkalaci: @chhetripradeep now pushed the following flow:

> Disable cache
Run the first execution of all queries
Enable cache & Synchronously load the data
Run the second and the third executions of the queries

onderkalaci: Hi @chhetripradeep,

Just to confirm, I'm assuming the next steps are with you. I'm waiting for any further feedback you might have.

chhetripradeep: Hi @onderkalaci yes it's pending on me. Sorry for the delay. I will work on it in 1-2 days.

crunchy-bridge-for-analytics/results/crunchy-bridge-analytics-256.json
```diff
@@ -0,0 +1,55 @@
+{
+    "system": "Crunchy Bridge for Analytics (Parquet)",
+    "date": "2024-06-05",
+    "machine": "Crunchy Bridge (Analytics-256GB)",
```
rschu1ze, line 4: The "Machine" facet at https://benchmark.clickhouse.com/ usually refers to EC2 instance types or a `[CPU, RAM]` pair (except for t-shirt sizes 'M', 'S' etc. for ByteHouse and Snowflake who don't reveal the underlying hardware. "Analytics-256" does not immediately make the used hardware clear. I went through the steps described in README.txt of this PR and find that at some point "64 vCores / 256 GB DRAM" was shown for "Analytics-256". Could we therefore write `"machine": "Analytics-256GB (64 vCores, 256 GB)"` in l. 4?

crunchy-bridge-for-analytics/README.md
```diff
@@ -0,0 +1,46 @@
...
+[Crunchy Bridge for Analytics](https://docs.crunchybridge.com/analytics) allows users to query parquet files on S3 directly.
+
+
+1) Setup the cluster:
+
+- Go to https://crunchybridge.com/
+- Click `Create Cluster` -> `Create Analytics Cluster`
+- Choose Region (`eu-central-1`) where `s3://clickhouse-public-datasets/hits_compatible/hits.parquet` is
+- Click `Analytics-256`
+- Click `Create Cluster`
```
rschu1ze, line 11: (Tried to reproduce using the steps but the UI wanted my credit card details first to charge me > 6.8k $.)

crunchy-bridge-for-analytics/run.sh
```diff
@@ -0,0 +1,57 @@
+#!/bin/bash
+
+# disable cache manager to prevent auto-loading of files
+$SUPERUSERCONNCMD -c 'ALTER SYSTEM SET crunchy_query_engine.enable_cache_manager TO off;'
```
rschu1ze, line 4: Just for my curiosity: Is this comment somewhere documented? At least Google does not know about it: https://www.google.com/search?q=crunchy_query_engine+enable_cache_manager&client=firefox-b-d&sca_esv=9267af3241730e66&ei=586sZpDcH52K9u8Ph-KTqAw&ved=0ahUKEwjQ6_DApdaHAxUdhf0HHQfxBMUQ4dUDCBE&uact=5&oq=crunchy_query_engine+enable_cache_manager&gs_lp=Egxnd3Mtd2l6LXNlcnAiKWNydW5jaHlfcXVlcnlfZW5naW5lIGVuYWJsZV9jYWNoZV9tYW5hZ2VyMggQABiABBiiBEiXCFCFBljdBnABeACQAQCYAZkBoAGJAqoBAzAuMrgBA8gBAPgBAZgCAaACmgGYAwCIBgGSBwMwLjGgB-oF&sclient=gws-wiz-serp

crunchy-bridge-for-analytics/results/crunchy-bridge-analytics-256.json
onderkalaci: sure, I can update this.

crunchy-bridge-for-analytics/create.sql
rschu1ze: The rules of ClickBench:

```
# Caching

If the system contains a cache for query results, it should be disabled.

It is okay if the system performs caching for source data (buffer pools and similar). If the cache or buffer pools can be flushed, they should be flushed before the first run of every query.

If the system contains a cache for intermediate data, that cache should be disabled if it is located near the end of the query execution pipeline, thus similar to a query result cache.
```

For me, it sounds like the rules were made before Cloud-native databases existed, i.e. they don't consider local caching of remote object store files. Anyways, either viewpoint is valid, I would say. We can go with the current approach (cold = remote access to S3, warm = read from locally downloaded files), but perhaps the README.md could point out more clearly what happens.

crunchy-bridge-for-analytics/results/crunchy-bridge-analytics-256.json
rschu1ze: Thinking about it, results would be more comparable overall if they were taken on a smaller machine, e.g. something equivalent to c6a.4xlarge (only if this is possible).

crunchy-bridge-for-analytics/README.md
onderkalaci: Well, that's for full month:) The costs are accrued as a pro rata portion of the monthly value, on minute basis.

(And, you can also [suspend](https://docs.crunchybridge.com/how-to/suspend-resume) the cluster when not used.)

crunchy-bridge-for-analytics/run.sh
onderkalaci: I think it is not documented in our docs. 

In practice, no user would ever really need this. On all the instance types that we provide on Crunchy Bridge for Analytics, the local file caching is always enabled. That's probably why we haven't documented it (yet).

crunchy-bridge-for-analytics/create.sql
onderkalaci: >  but perhaps the README.md could point out more clearly what happens.

Sure, let me update the readme

rschu1ze (approved):

PR #201, opened 2024-06-17
Revert "SelectDB update results on metal machine"

chhetripradeep:
This reverts commit 9c946f7129ed2b412d325490ec26933f3eef6d88.

Hi @Gabriel39 

Unfortunately we have to revert [your earlier PR](https://github.com/ClickHouse/ClickBench/pull/185). We found there are few things which goes against the ClickBench rules. We would request you to fulfil the following ClickBench requirements and resubmit the PR again. I will be very happy to review your updated PR again.

#### Query Result Caching

Regarding [Caching](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#caching), ClickBench says "If the system contains a cache for query results, it should be disabled."

In our testing we found the SelectDB is doing query result caching.

First run of the query returns in `2.5 sec`
```
mysql> select * from hits where FUniqID = 8662277244489081482;
--------------
select * from hits where FUniqID = 8662277244489081482
--------------
...
7 rows in set (2.57 sec)
```
Second run of the query return quite fast in `0.03 sec` since it is coming from cache.
```
mysql> select * from hits where FUniqID = 8662277244489081482;
--------------
select * from hits where FUniqID = 8662277244489081482
--------------
...
7 rows in set (0.03 sec)
```
This cache doesn't get dropped after dropping linux file cache and subsequent runs of the query return quite fast from cache in `0.03 sec` 
```
$ echo 3 | sudo tee /proc/sys/vm/drop_caches
3

mysql> select * from hits where FUniqID = 8662277244489081482;
--------------
select * from hits where FUniqID = 8662277244489081482
--------------
...
7 rows in set (0.03 sec)
```
Only way was to drop the cache was to restart the fe and be servers

```
$ ./selectdb/output/fe/bin/stop_fe.sh
$ ./selectdb/output/be/bin/stop_be.sh
$ ./selectdb/output/fe/bin/start_fe.sh --daemon
$ ./selectdb/output/be/be/bin/start_be.sh --daemon

mysql> select * from hits where FUniqID = 8662277244489081482;
--------------
select * from hits where FUniqID = 8662277244489081482
--------------
...
7 rows in set (3.56 sec)
```

#### Wrong Query Response

SelectDB returned wrong results for two queries: Q14 and Q29

Regarding Q14, we found that SelectDB return different results in every run. 
- You can see the result of 3 consecutive SelectDB runs here: https://pastila.nl/?00caa21b/dc0ff1354940097dc8d1c14b49edfe8a#6BaDeFUDkwe098OEE4W7BQ==
- While other databases gives same result in every run. You can see the result of 3 consecutive of ClickHouse runs here: https://pastila.nl/?0066d7be/9cf8222c7879f25a6528278482f4100e#V3YJ4CiNHzoOVvgH/NInLQ== 
- In order to do extra check, we even ran the query against DuckDB. You can see the result of 3 consecutive of DuckDB runs here: https://pastila.nl/?00253def/bcf2e1963231b230b1b56e613cb0f2dd#YZoCkSD/mMtx9vZ1hnnChQ==

Regarding Q29, we found similar issue where SelectDB is returning different result. You can see the result of 3 databases here: https://pastila.nl/?002f5166/24cd8f2b8e9bf08d5ecdfc31d38075ae#JRkbasaN4cgHG6ZMnhE+jw==

Overall results from ClickHouse and DuckDB matches and is consistent across different runs but results from SelectDB is different and is inconsistent in every run.

#### Specialized Build 

We saw that you used some specialized builds like `https://doris-clickbench.oss-ap-southeast-1.aliyuncs.com/selectdb-linux_x64.tar.gz` We would like all databases to use the standard open source binary. Hence we would like you to update that too.


alexey-milovidov (approved):

Gabriel39:
Hi @chhetripradeep @alexey-milovidov , thanks for your attention.

I want to figure out two things which are confused for me.

1. For query `select * from hits where FUniqID = 8662277244489081482;`, why do you think it hits a query cache? To be honest, I think results returned by query cache will be far sooner than `~0.03 sec` now. To prove this, I compared this query's profile between cold run and hot run. And then, I noticed that the only different is the hot run 's data scanning hit SelectDB's built-in data page cache and cold run did not. I think a built-in data page cache is very common for mainstream data warehouse system and it **does not** go against the ClickBench [rules.]
[cold run.txt](https://github.com/user-attachments/files/15866136/cold.run.txt)
[hot run.txt](https://github.com/user-attachments/files/15866145/hot.run.txt)
2. For Q29, I tested on **[DuckDB 1.0.0](https://github.com/duckdb/duckdb/releases/download/v1.0.0/duckdb_cli-linux-amd64.zip)** which is claimed as a stable version. To assure the results accurate, I did not load data into DuckDB and just executed Q29 on DuckDB by an `external-table way`(`SELECT REGEXP_REPLACE(column014, '^https?://(?:www\.)?([^/]+)/.*$', '\1') AS k, AVG(STRLEN(column014)) AS l, COUNT(*) AS c, MIN(column014) FROM 'hits.tsv'  WHERE column014 <> '' GROUP BY k HAVING COUNT(*) > 100000 ORDER BY l DESC LIMIT 1;`). And then , I found results returned by DuckDB is the same as SelectDB and different from ClickHouse.
image


And for Q14 you mentioned, I think it must be something wrong in SelectDB. Frankly speaking, this binary is based on a developing version which may not be stable enough.

PR #209, opened 2024-07-29
Add Dell r660

rschu1ze:
Results provided by Sergii Batishchev. Thanks!


Slach:
hm, looks weird, PR merged but https://benchmark.clickhouse.com/hardware/ doesn't show R660

Issue #119, opened 2023-06-13
Add YTsaurus support


zamazan4ik:
What is YTsaurus: https://ytsaurus.tech/

Since YTsaurus [does not have](https://github.com/ytsaurus/ytsaurus/issues/21#issuecomment-1488384350) a built-in benchmark tool, could be a good idea to add YTsaurus support to ClickBench to get a possibility to benchmark YTsaurus somehow. Since YTsaurus supports YQL and has CLI, it should be possible quite easy to adopt ClickBench (at least from the interface perspective).

Honestly, not sure how typical workloads from ClickBench are for YTsaurus - needs to be clarified.

@gritukan you could be interested in this issue. Resolving this one can help with https://github.com/ytsaurus/ytsaurus/issues/40

alexey-milovidov:
Done.

PR #224, opened 2024-09-16
Add CHYT 

sesho96:
Resolves: #119

This PR adds CHYT (ClickHouse over YTsaurus) results and benchmark scripts for ClickBench


chyt/README.md
```diff
@@ -0,0 +1,14 @@
+#### CHYT powered by ClickHouse
```
rschu1ze, line 1: "CHYT" ... what does that mean? Should it simply be "YTsaurus"?

chyt/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+1.  Install YTsaurus cluster. Visit [YTsaurus Getting started webpage](https://ytsaurus.tech/docs/en/overview/try-yt)
```
rschu1ze, line 3: As far as I understand, the code is open-source, right? Script `benchmark.sh` ideally does as much setup as possible automatically, i.e. no user intervention. For examples how to do that, please see clickhouse/benchmark.sh, postgresql/benchmark.sh and duckdb/benchmark.sh.

chyt/results/yt.192GB_YC.json
```diff
@@ -0,0 +1,60 @@
+
+{
+    "system": "CHYT",
+    "date": "2024-09-16",
+    "machine": "192GB",
```
rschu1ze, line 5: I am confused. L. 6 says "serverless" which typically means the results were measured in a database-as-a-service offering (such as ClickHouse Cloud). Was that the case?

If not, it would be good to specify the exact machine specs for reproducibility, see e.g. duckdb/results/c5.4xlarge.json.

rschu1ze (commented):
Thanks for the PR.

It would be great if the benchmark script could be more "automatic", i.e. download the database, configure it, start it, import the data, and run the scripts without user intervention. It will make it much easier for "outsiders" like me to reproduce the results.

chyt/README.md
sesho96: YTsaurus is a data platform. CHYT is high-level computing engine inside YTsaurus [architecture](https://ytsaurus.tech/docs/en/overview/about).

chyt/README.md
sesho96: Yes, YTsaurus is open-source system. But CHYT is a small part of it. `Benchmark.sh` uses pre-installed cluster with default clique to do benchmark test.
YTsaurus cluster can be installed, for example, using [k8s operator](https://github.com/ytsaurus/ytsaurus-k8s-operator)
All possible variants are described in [documentation](https://ytsaurus.tech/docs/en/overview/try-yt).

chyt/results/yt.192GB_YC.json
sesho96: For CHYT with 48, 96 and 192 GB we use 1, 2 and 4 instances with 12 vCPU and 48 Gb RAM
For CHYT with 360 and 720 GB  -- 9 and 18 instances with 10 vCPU and 40 Gb RAM

You only configure the count and size of instances and then YTsaurus will schedule them across computational nodes of cluster.

chyt/README.md
rschu1ze: As I mentioned, we need to reduce the variability here ... As someone who wants to verify the benchmark results, I like to run `benchmark.sh` and have it install everything by itself. The only think that I would be able to choose is the hardware the system runs on.

sesho96: We have a possibility to create a [demo cluster](https://ytsaurus.tech/docs/en/overview/try-yt#demo-stand) for everyone who wants to try YTsaurus. Would it be ok to verify the results?

rschu1ze: One installation option is [Docker](https://ytsaurus.tech/docs/en/overview/try-yt#using-docker). That seems the alternative with the least amount of complexity and the best reproducibility (compared to k8s and the demo cluster).

My preference would be if benchmark.sh sets up the docker container, does other preparations, and then runs the measurements.

sesho96: Deploying a cluster with 360 vCPUs and 720 GB of RAM using a single VM can be quite challenging. In our case, we use a Kubernetes cluster with nodes of the type c6a.8xlarge and network SSDs that perform similarly to gp2 volumes.
We also aim to demonstrate various cluster sizes, not just the smallest one. If necessary, we can remove the "serverless" tag and instead specify the number of CHYT instances in the configuration.
Given the large size of our cluster, Docker deployment was not utilized for benchmarking, as it may yield different results. The easiest way to reproduce our results is by booking a demo cluster through our website.
Additionally, I can include a step-by-step guide in the README.md file to assist with the setup.

rschu1ze: Okay, let's add a step by step guide in the README, I'll afterwards try my best to reproduce, then I will merge.

PR #235, opened 2024-10-30
Update MotherDuck results, switch to parquet file for loading.

jtigani:
MotherDuck results were quite old, so I updated to recent versions. Also data was being loaded from a compressed csv file, which is really inefficient because it can't be parallelized. So I switched to read from parquet (note that the parquet file doesn't have the timestamps correctly represented so they need to be coerced).


jtigani:
@rschu1ze PTAL.

jtigani:
Thanks Robert ... Do you know what region you ran from? Since this is a cloud service that is currently singly-hosted in US-EAST1, it is pretty sensitive to location you launch the benchmark from. The numbers I got were running from an EC2 instance in US-EAST1 to have it be close to our servers (we're adding more regions soon). But if you run the benchmark from other locations you end up spending a lot of time in transit to our cloud service. Eyeballing your numbers it looks like if you ran from another region that would account for a lot of the difference.

(also if you're using the free tier you may get throttled, but that doesn't seem to be happening because your queries look like they just have an extra ~100ms added to them).

rschu1ze:
I ran the script from a server in Europe, that might explain it. Thanks, I'll merge then!

motherduck/results/result.json
```diff
@@ -1,57 +1,57 @@
...
+[0.19063701599952765,0.18687298700024257,0.18603799500124296],
+[0.21010683900021832,0.1866715190008108,0.198137998002494],
+[0.38373803200011025,0.37373249000302167,0.371793443999195],
+[0.3542730239969387,0.33429408599840826,0.33363083899894264],
+[0.6366427199973259,0.6410418160012341,0.6345498260016029],
+[0.05052374699880602,0.014082836998568382,0.014972674998716684],
+[0.3299962970013439,0.30164447399874916,0.3019338140002219],
+[0.2213447880021704,0.221652765001636,0.22077769899988198],
+[0.3728656390012475,0.3468815829983214,0.34391914299703785],
+[1.2559020400003647,1.2022961519978708,1.2722901870001806],
```
rschu1ze, line 36: @jtigani Just for my understanding: Cold (1st) and hot (2nd/3rd) runtimes are quite similar in most of the new measurements. Given that the `hits` table is quite big (75GB TSV), this can mean a few things:
- the servers are equipped with lots of DRAM and the data remains essentially in memory after the first load. I am speculating of course but DRAM is quite costly so I'd deem that unlikely.
- the data is compressed so well that it fits into the main memory of a smaller server
- scans are avoided using skipping techniques. I [recently tried](https://github.com/ClickHouse/ClickHouse/issues/70605#issuecomment-2457022442) min-max (SMA/zonemap) indexes recently on ClickBench but it did not make a difference. (but of course, there other skipping techniques exist, so this is not representative).

No need to reveal business secrets :) , I just wanted to make sure that the measurements contain no accidental mistake. Thanks!

PR #246, opened 2024-11-12
Added tinybird

melvynator:
Added Tinybird (default experience) benchmark.


alexey-milovidov (approved):

javisantana:
did you run the benchmark on a free account?

melvynator:
Yes. Also worth noting I've requested to increase the timemout but it was suggested to rewrite the query instead which I could not do fairly, so that's why some of this query did not finish.

alexey-milovidov:
We allow both free and commercial accounts. The only requirement is that is has to be reproducible in a self-service manner, so: - without custom configurations; - without asking for support.

javisantana:
That’s totally fine, if you want to run longer queries you can use
asynchronous api but I guess that’s outside of the scope of the benchmark.

On top of that raising the timeout would break the rules (as it’s custom
config)

We of course knew you were running the test, we could have routed your
workspace to a massive cluster but decided to play fair. I guess explaining
the benchmark was ran in a free account would be also fair.

Thanks for running the test

melvynator:
Done: https://github.com/ClickHouse/ClickBench/pull/248

melvynator:
Also added to the MD.

PR #250, opened 2024-11-14
Add Results for Timescale Cloud

matthewmcnew:
- Uses Terraform to setup Timescale Cloud instance
- Results for 4cpu/8cpu/16cpu instances


timescale-cloud/benchmark.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# This benchmark uses terraform to provision an Timescale Cloud instance. https://www.timescale.com/blog/create-timescale-services-with-terraform-provider/
+# export TS_PROJECT_ID=...
+# export TS_ACCESS_KEY=...
+# export TS_SECRET_KEY=...
+# export REGION=${REGION:-us-east-2}
+# export MILLI_CPU=${MEMORY:-32000}
+
+#install postgres client
```
rschu1ze, line 10: For my understanding: Are Timescale Cloud users also required to provision an instance via Teraform? I guess not?

While this script seems very much automated, this is technically not even necessary for Cloud databases. It is okay to describe the steps to create an account via the web UI, and then connect to the instance programmatically.

Something similar was done for Tinybird recently: https://github.com/ClickHouse/ClickBench/tree/main/tinybird

Such an approach is actually more realistic in the sense that it captures the experience of a new user with the cloud database service.

matthewmcnew: Thanks for the feedback @rschu1ze, I updated the PR with a README that walks through the benchmark process.

rschu1ze (approved):

PR #253, opened 2024-11-15
pg_duckdb

serprex:
This queries local parquet files, a `pg_duckdb-motherduck` variant can be added in future using motherduck integration instead of `read_parquet`


pg_duckdb/benchmark.sh
```diff
@@ -0,0 +1,17 @@
...
+
+set -e
+
+sudo apt-get update
+sudo apt-get install -y docker.io
+sudo apt-get install -y postgresql-client
+
+sudo docker run -d --name pgduck -p 5432:5432 -e POSTGRES_PASSWORD=duckdb pgduckdb/pgduckdb:16-main
+
+psql postgres://postgres:duckdb@localhost:5432/postgres -f create.sql
```
serprex, line 11: container isn't immediately available, need to have a wait loop

pg_duckdb/run.sh
```diff
@@ -0,0 +1,13 @@
...
+TRIES=3
+
+cat queries.sql | while read -r query; do
+    sync
+    echo 3 | sudo tee /proc/sys/vm/drop_caches
+
+    echo "$query";
+    for i in $(seq 1 $TRIES); do
+        psql postgres://postgres:duckdb@localhost:5432/postgres -t -c '\timing' -c "$query" | grep 'Time'
+    done;
```
JelteF, line 12: Opening and closing the postgres connection for each query has significant overhead when connecting to MotherDuck, because each postgres backend needs to open its own connection to motherduck. Writing a Python script instead might be preferable, otherwise you can do something like this.

```suggestion
    echo "$query"
    (
        echo '\timing'
        yes "$query" | head -n $TRIES
    ) | psql postgres://postgres:duckdb@localhost:5432/postgres -t | grep 'Time'
```

pg_duckdb/create.sql
```diff
@@ -0,0 +1,2 @@
+SELECT duckdb.cache('s3://clickhouse-public-datasets/hits_compatible/hits.parquet', 'parquet');
```
JelteF, line 1: Instead of caching the S3 file you should also be able to use a local file in the view like this:

```sql
SELECT * FROM read_parquet('/tmp/hits.parquet') AS (...)
```

JelteF (commented):
> TODO: should we split up another variant between pg-duckdb-motherduck & pg-duckdb-parquet

Yeah, I think that makes sense. I/we can contribute the pg-duckdb-motherduck one if needed.

> TODO: test, view probably doesn't work, but seems weird inlining huge schema on every query. can have some sed replace on \bhits\b I guess

Using a view should work fine, that's currently the recommended approach when reading from the same parquet file multiple times.

pg_duckdb/run.sh
JelteF: Other than that, for MotherDuck it's important that the machine that runs the benchmark is in AWS us-east-1 to get the lowest latency to MotherDuck.

pg_duckdb/create.sql
serprex: done. mounting greatly sped things up compared to using cache

pg_duckdb/benchmark.sh
serprex: decided to be lazy & went with `sleep 5` because it works

Issue #268, opened 2024-11-25
Polars benchmark are very suboptimal.


ritchie46:
They benchmark the eager API which forces Polars to materialize every result and doesn't allow any optimizations. We state in our docs that if you benchmark Polars you should use the Lazy API. 

I also wonder why data must be loaded via Pandas?

I could take a stab at doing correct Polars queries. Does IO time need to be included or must queries load from DataFrames. I also wonder why pandas queries are mixed with the Polars queries file?

rluvaton:
There should also be polars from parquet and sql like in datafusion and use [`polars-cli`](https://github.com/pola-rs/polars-cli) for that

auxten:
@ritchie46 
For this patch, it is a test of 'Query on Dataframe'. As polars do not support query on dataframe directly, there is a load time.

You are welcomed to send a patch to correct usage of Polars😀

ritchie46:
> As polars do not support query on dataframe directly, there is a load time.

I don't know what you mean? What do we not support?

ritchie46:
> There should also be polars from parquet and sql like in datafusion and use [`polars-cli`](https://github.com/pola-rs/polars-cli) for that

We can write the queries from parquet. I can adapt the PR to do that. Our polars-cli is heavily outdated. I would recommend using Python Polars. That's our latest binary. We also can run SQL (but not everything as SQL is more an afterthought).

auxten:
> > As polars do not support query on dataframe directly, there is a load time.
> 
> I don't know what you mean? What do we not support?

I mean for #222, it a "Query on Pandas Dataframe benchmark".
Forgive my ignorance, I don't know how Polars can query on a Pandas DataFrame directly without converting it to a Polars DataFrame.

ritchie46:
It is not meant to query on pandas DataFrames. If pandas can query on it's own Dataframes, I think Polars should be able to do that as well.

Issue #264, opened 2024-11-22
Polars vs DataFusion benchmark results


rluvaton:
I would like to have polars vs data fusion results, but they have no common machine that I can compare with

What does it take to add it?

rschu1ze:
You will need to set up an AWS EC2 machine with the desired instance type (see [here](https://aws.amazon.com/de/ec2/instance-types/)). During setup, go with the default.

Then clone the ClickBench repository and run polars/benchmark.sh resp. datafusion/benchmarks.sh.

alexey-milovidov:
@rschu1ze, please see https://github.com/ClickHouse/ClickBench/pull/270#issuecomment-2543444463

PR #270, opened 2024-11-25
Fix polars

ritchie46:
Hey, I am from the Polars project and was surprised Polars was added to ClickBench. When I looked at the implementation I saw that Polars was incorrectly used (#268). and sometimes even python functions were called where we have a native API for the operations.

This ensures that 

- Polars' queries can use the Lazy API and thus the optimizer and the query engine. 
- The statistics are computed in single query (similar to SQL). That ensures we can parallelize
- Use Polars expressions and not `map_elements` with python udfs

Resolves: https://github.com/ClickHouse/ClickBench/issues/268


polars/query.py
```diff
@@ -22,48 +22,48 @@
...
         "SELECT COUNT(*) FROM hits WHERE AdvEngineID <> 0;",
         lambda x: x[x["AdvEngineID"] != 0].count(),
-        lambda x: x.filter(pl.col("AdvEngineID") != 0).height,
+        lambda x: x.select(pl.col("AdvEngineID").filter(pl.col("AdvEngineID") != 0).count()).collect().height,
     ),
     (
         "Q2",
         "SELECT SUM(AdvEngineID), COUNT(*), AVG(ResolutionWidth) FROM hits;",
         lambda x: (x["AdvEngineID"].sum(), x.shape[0], x["ResolutionWidth"].mean()),
-        lambda x: (x["AdvEngineID"].sum(), x.height, x["ResolutionWidth"].mean()),
```
ritchie46, line 42: These are 2 queries, where in SQL it is in a single query.

polars/query.py
```diff
@@ -22,48 +22,48 @@
...
+        lambda x: x.select(pl.col("advengineid").sum(), pl.len(), pl.col("advengineid").mean()).collect().rows()[0],
     ),
     (
         "Q3",
         "SELECT AVG(UserID) FROM hits;",
         lambda x: x["UserID"].mean(),
-        lambda x: x["UserID"].mean(),
```
ritchie46, line 48: This is not lazy suitable. Forces a single engine. It is meant for notebooks.

polars/query.py
```diff
@@ -318,15 +318,13 @@
         .group_by("CounterID")  # GROUP BY CounterID
         .agg(
             [
-                pl.col("URL")
-                .map_elements(lambda y: len(y), return_dtype=pl.Int64)
```
ritchie46, line 322: This was wrong. Didn't compute the `AVG` and it uses a python function, where we have native expressions for this.

polars/query.py
```diff
@@ -352,18 +350,14 @@
             .group_by("k")
             .agg(
                 [
-                    pl.col("Referer").map_elements(
-                        lambda y: len(y), return_dtype=pl.Int64
```
ritchie46, line 356: This was wrong. Didn't compute the `AVG` and it uses a python function, where we have native expressions for this.

ritchie46:
Yes, current code uses pandas to read parquet. I don't know why?

The error you see is a pandas installation error. Shall I remove pandas from that script?

rschu1ze:
> Yes! and if you can run the benchmark c6a.4xlarge, 500gb gp2 (like in datafusion) it would be great!

This relates to #264. I actually tried to run polars/benchmark.sh on my corporate c6a.4xlarge machine (using the version in this PR, now that [the setup is fixed](https://github.com/ClickHouse/ClickBench/pull/270#issuecomment-2499660384)) ... and it froze my machine so badly that it did not come back in 2.5 hours after which I killed the instance forcefully. Perhaps someone else has more luck.

ritchie46:
> and it froze my machine so badly that I was unable to connect via SSH or ping it for 2.5 hours,

Yes, this version loads all data in memory which doesn't fit that machine. This was already the case, and I didn't adapt that in the PR.

I can also follow up with a PR where we load data from parquet. That doesn't materialize all data and is much less memory intensive.

rschu1ze:
> I can also follow up with a PR where we load data from parquet. That doesn't materialize all data and is much less memory intensive.

That would be nice. Actually, we should try to do changes to the benchmark scripts / queries and the results themselves in the same PR. Otherwise, one of them will be outdated. So I guess my preference is if we could update all polars variants within this PR (and if it helps, I can setup myself a `c6a.metal` box which is _way_ more powerful than `c6a.4xlarge` and try to redo the measurements).

rschu1ze:
@ritchie46 Side question: Like to sign above CLA declaration as well? I don't care too much personally but the lawyers might. Thanks.

ritchie46:
> @ritchie46 Side question: Like to sign above CLA declaration as well? I don't care too much personally but the lawyers might. Thanks.

Signed. :+1:

ritchie46:
@rschu1ze I have added parquet as a source as well. Made it so that we create 2 output jsons, indicating the source used.

Note that running from parquet has drastically reduced memory usage. The DataFrame case is just very intensive as we are forced to load all data in memory.

rschu1ze:
@ritchie46 Thanks ... I'll re-run measurements in the coming days!

ritchie46:
Great, thanks!

rschu1ze:
I both dataframes and parquet versions on c6a.4xlarge again but the system hang. Didn't check deeper why, instead I repeated the measurements on a c6a.metal machine. Pushed the measurements into this PR.

alexey-milovidov:
@rschu1ze, the server's freeze means OOM. Please run it with `prlimit` as here:
https://github.com/ClickHouse/ClickBench/pull/267/files#diff-6ad3e2e988a332b62ba8cc145121a402e3372f9414784d7c9a93e7b6967e8be6R11

Some queries will be shown as failed, but the results are still good to publish (and then they can be improved).

rschu1ze:
Done: https://github.com/ClickHouse/ClickBench/pull/279

PR #271, opened 2024-11-25
Update results for QuestDB 8.2

puzpuzpuz:
We switch to partitioned table on the c6a.4xlarge instance type, so now all results are in `./questdb` directory.


rschu1ze (approved):

puzpuzpuz:
@rschu1ze thanks for the review!

PR #279, opened 2024-12-15
Add Polars (Parquet) on `c6a.4xlarge`

### LINKED ISSUE #264: Polars vs DataFusion benchmark results

rschu1ze:
Resolves #264

https://github.com/ClickHouse/ClickBench/pull/270#issuecomment-2543444463


alexey-milovidov:
@rschu1ze, ok, but I don't see `prlimit` in the automated script. It makes the results not reproducible.

PR #284, opened 2025-01-10
add postgres (with pg_mooncake columnstore)

zhousun:
Add results for postgres with extension pg_mooncake (columnstore table in postgres)


pg_mooncake/benchmark.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+
+sudo apt-get update
+sudo apt-get install -y docker.io
+sudo apt-get install -y postgresql-client
+
+wget --no-verbose --continue https://datasets.clickhouse.com/hits_compatible/athena/hits.parquet
+sudo docker run -d --name pg_mooncake -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust - -v ./hits.parquet:/tmp/hits.parquet mooncakelabs/pg_mooncake
```
rschu1ze, line 9: Do we really need to run docker with root permissions?

rschu1ze, line 9: L. 9 gives me `docker: invalid reference format.`.

What worked for me was

```
docker run -d --name pg_mooncake -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -v ./hits.parquet:/tmp/hits.parquet -- mooncakelabs/pg_mooncake
```

Kindly double-check.

pg_mooncake/benchmark.sh
```diff
@@ -0,0 +1,22 @@
...
+
+sleep 5
+psql postgres://postgres:pg_mooncake@localhost:5432/postgres -f create.sql
```
rschu1ze, line 12: This returned

```
65b913e46da5fd63729ffcd82a3d8cc92c50012077a4b9180081ef5e1d7414a2
psql:pg_mooncake/create.sql:108: ERROR:  access method "columnstore" does not exist
psql:pg_mooncake/create.sql:112: ERROR:  relation "hits" does not exist
LINE 1: insert into hits select WatchID, JavaEnable, Title, GoodEven...
```

At that point, I stopped checking further. Could you please fix these issues?

zhousun:
@rschu1ze thanks for review. Sorry some CI on my end.docker was not picking the correct version and forgot to create the extension :(  
Everything should work now.

pg_mooncake/benchmark.sh
Felixoid: +1 to get rid of `sudo` parts and make all preparations in the running container.

Let's assume that docker is installed on a host. And we should not stick to the Debian ecosystem.

pg_mooncake/results/c6a.4xlarge.json
```diff
@@ -0,0 +1,58 @@
+{
+    "system": "PostgreSQL with pg_mooncake",
+    "date": "2025-01-08",
+    "machine": "c6a.4xlarge, 500gb gp2",
+    "cluster_size": 1,
+    "comment": "",
+
+    "tags": ["C", "columnstore-oriented", "PostgreSQL compatible"],
```
rschu1ze, line 8: No existing database in ClickBench is tagged as `columnstore-oriented`, but many existing databases are tagged as `column-oriented`.

Let's rename that tag.

pg_mooncake/run.sh
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash
+
+TRIES=3
+CONNECTION=postgres://postgres:pg_mooncake@localhost:5432/postgres
+
+cat queries.sql | while read query; do
+    echo "$query"
+    (
+        echo '\timing'
+        yes "$query" | head -n $TRIES
```
rschu1ze, line 10: As per the [benchmark rules](https://github.com/ClickHouse/ClickBench), `[...] the first query should be run after dropping the page cache [...]`. Please grep for `/proc/sys/vm/drop_caches` in the ClickBench repository for examples how to do that.

pg_mooncake/benchmark.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+
+sudo apt-get update
+sudo apt-get install -y docker.io
+sudo apt-get install -y postgresql-client
+
+wget --no-verbose --continue https://datasets.clickhouse.com/hits_compatible/athena/hits.parquet
+sudo docker run -d --name pg_mooncake -p 5432:5432 -e POSTGRES_HOST_AUTH_METHOD=trust -v ./hits.parquet:/tmp/hits.parquet mooncakelabs/pg_mooncake:17-v0.1.0
```
rschu1ze, line 9: Still using `sudo` (also in l. 19), is this really necessary?

pg_mooncake/results/c6a.4xlarge.json
```diff
@@ -0,0 +1,58 @@
...
+
+    "load_time": 576,
+    "data_size": 14619781667,
+
+    "result": [
```
rschu1ze, line 13: I was able to reproduce these measurements locally 👍

zhousun:
updated. thx for the review

pg_mooncake/run.sh
rschu1ze: Let's address this ^^ as well.

pg_mooncake/benchmark.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+
+#install docker/ postgres if needed.
+
+# sudo apt-get update
```
rschu1ze, line 6: Fresh benchmark systems will not have postgresql-client. Maybe keep l. 6-8 in?

PR #285, opened 2025-01-12
Updated Opteryx results (v0.19.1)

joocer:
Rerun of the benchmark against the latest version (0.19.1)

Following the initial successful run of Opteryx against the benchmark, it was clear there were inefficiencies, the largest of which was due to the use of the pooling and buffering of reads intended for high-latency S3/GCS stores even though the files in the benchmark are on low-latency local storage. 

0.19.0 to 0.19.1 largely focused on improving Local IO, a feature that existed primarily for offline development and CI tests and hadn't been a focus for performance as the core user group uses S3/GCS as their storage.

Regards @Joocer


opteryx/README.md
```diff
@@ -1,30 +1,59 @@
...
+4. Run the benchmark script.
+
+### **Detailed Instructions**
+
+1. **Start an AWS EC2 instance**
+   - OS: Ubuntu 24
+   - Architecture: 64-bit
+   - Instance Type: `c6a.4xlarge`
+   - Root Storage: 500 GB gp2 SSD
+   - Advanced Settings: Ensure EBS-optimization is **disabled**.
```
rschu1ze, line 31: I see this when I launch an EC2 instance:

image

What does "EBS-optimization" mean and where do I find it in their GUI?

rschu1ze (approved):

PR #288, opened 2025-01-18
Hyper results with version 0.0.21200

caetanosauer:
This PR adds measurements for the [Hyper](https://tableau.github.io/hyper-db/) database using our Python API that is installed via pip.

The measurements include both our native format as well as the provided partitioned parquet files, on both c6a.metal and c6a.4xlarge.

Please let me know if there is any issue with the submission and I'd be happy to readjust. Thank you very much for your work on this benchmark and this valuable contribution to the community!

Caetano


hyper-parquet/benchmark.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+
+sudo apt-get update
+sudo apt-get install -y python3-pip
+pip install tableauhyperapi
```
rschu1ze, line 5: On a default Ubuntu 24.04 in a default c6a.4xlarge EC2 instance, I got this error:

```
error: externally-managed-environment

× This environment is externally managed
╰─> To install Python packages system-wide, try apt install
    python3-xyz, where xyz is the package you are trying to
    install.

    If you wish to install a non-Debian-packaged Python package,
    create a virtual environment using python3 -m venv path/to/venv.
    Then use path/to/venv/bin/python and path/to/venv/bin/pip. Make
    sure you have python3-full installed.

    If you wish to install a non-Debian packaged Python application,
    it may be easiest to use pipx install xyz, which will manage a
    virtual environment for you. Make sure you have pipx installed.

    See /usr/share/doc/python3.12/README.venv for more information.

note: If you believe this is a mistake, please contact your Python installation or OS distribution provider. You can override this, at the risk of breaking your Python installation or OS, by passing --break-system-packages.
hint: See PEP 668 for the detailed specification.
```

Adding `--break-system-packages` made pip work. Like to add this flag?

caetanosauer: Hi Robert,

Thanks so much for looking at our PR and spotting the problem. We used Ubuntu 22.04 where PEP 668 was not an issue yet. Apologies for the oversight. I added the flag in the latest commit.

rschu1ze:
@caetanosauer Thanks for the PR. I was able to re-produce the standard and Parquet results on a c6a.4xlarge machine. Merging!

PR #302, opened 2025-02-05
[DuckDB] Update DuckDB results to DuckDB v1.2.0 and switch to using the CLI

Mytherin:


rschu1ze:
I re-ran the standard configuration (non-Parquet) on a c6a.4xlarge instance and the measurements reproduced nicely. Thanks, merging.

Issue #234, opened 2024-10-24
Postgres-Tuned should tune alignment padding


meglio:
To make the tuned version of the postgres benchmark really tuned, you should optimize the columns layout, see "Column Tetris" at https://stackoverflow.com/a/7431468

alexey-milovidov:
Could you please edit the script? Then we will re-run it.

meglio:
Here's the `CREATE` statement with the improved order of columns:

```sql
CREATE TABLE hits
(
    WatchID BIGINT NOT NULL,
    UserID BIGINT NOT NULL,
    FUniqID BIGINT NOT NULL,
    ParamPrice BIGINT NOT NULL,
    RefererHash BIGINT NOT NULL,
    URLHash BIGINT NOT NULL,
    EventTime TIMESTAMP NOT NULL,
    ClientEventTime TIMESTAMP NOT NULL,
    LocalEventTime TIMESTAMP NOT NULL,
    EventDate Date NOT NULL,
    CounterID INTEGER NOT NULL,
    ClientIP INTEGER NOT NULL,
    RegionID INTEGER NOT NULL,
    RefererRegionID INTEGER NOT NULL,
    URLRegionID INTEGER NOT NULL,
    IPNetworkID INTEGER NOT NULL,
    SilverlightVersion3 INTEGER NOT NULL,
    CodeVersion INTEGER NOT NULL,
    HID INTEGER NOT NULL,
    RemoteIP INTEGER NOT NULL,
    WindowName INTEGER NOT NULL,
    OpenerName INTEGER NOT NULL,
    SendTiming INTEGER NOT NULL,
    DNSTiming INTEGER NOT NULL,
    ConnectTiming INTEGER NOT NULL,
    ResponseStartTiming INTEGER NOT NULL,
    ResponseEndTiming INTEGER NOT NULL,
    FetchTiming INTEGER NOT NULL,
    CLID INTEGER NOT NULL,
    JavaEnable SMALLINT NOT NULL,
    GoodEvent SMALLINT NOT NULL,
    CounterClass SMALLINT NOT NULL,
    OS SMALLINT NOT NULL,
    UserAgent SMALLINT NOT NULL,
    IsRefresh SMALLINT NOT NULL,
    RefererCategoryID SMALLINT NOT NULL,
    URLCategoryID SMALLINT NOT NULL,
    ResolutionWidth SMALLINT NOT NULL,
    ResolutionHeight SMALLINT NOT NULL,
[...]

PR #309, opened 2025-02-21
Refactor `postgresql/benchmark.sh` to Wrap COPY FREEZE in a Transaction Block

somratdutta:
## Overview
This PR fixes an issue in the PostgreSQL benchmark script where the use of the `COPY ... with freeze command` fails because PostgreSQL requires that the table be either created or truncated in the current subtransaction. 

Previously, the script executed the `\copy` command (with the freeze option) outside of any transaction block, resulting in the following error:
```shell
ERROR: cannot perform COPY FREEZE because the table was not created or truncated in the current subtransaction
```
## Changes

*   **Transaction Wrapping:**  
    The data load is now performed inside a transaction block. The script first truncates the `hits` table and then runs the `\copy` command with the `freeze` option—all within the same transaction. This guarantees that PostgreSQL accepts the `COPY FREEZE` operation.
  
Looking forward to your feedback and suggestions!

## Fix
This PR fixes this [issue](https://github.com/ClickHouse/ClickBench/issues/234#issuecomment-2499128995).

### DIFF
--- a/postgresql/benchmark.sh
+++ b/postgresql/benchmark.sh
@@ -10,7 +10,14 @@ chmod 777 ~ hits.tsv
 
 sudo -u postgres psql -t -c 'CREATE DATABASE test'
 sudo -u postgres psql test -t < create.sql
-sudo -u postgres psql test -t -c '\timing' -c "\\copy hits FROM 'hits.tsv' with freeze"
+
+# Load data: wrap TRUNCATE and \copy FREEZE in a single transaction
+sudo -u postgres psql test <<'EOF'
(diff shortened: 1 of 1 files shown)


somratdutta:
This PR fixes this https://github.com/ClickHouse/ClickBench/issues/234#issuecomment-2499128995.

rschu1ze:
@somratdutta I verified locally that your fix works (only made the comment a bit more verbose). Thanks a lot!

somratdutta:
Happy contributing 😄

PR #318, opened 2025-03-05
Update Hydra

wuputah:
This updates the Hydra benchmark to the latest version of the Hydra cloud service.


hydra/benchmark.sh
```diff
@@ -2,34 +2,11 @@
...
-
-# set up db, user, table; load data.
-sudo -u postgres psql -t -c 'CREATE DATABASE test'
-sudo -u postgres psql test -t < create.sql
-sudo -u postgres psql test -t -c 'TRUNCATE hits'
-sudo -u postgres psql test -t -c '\timing' -c "\\copy hits FROM 'hits.tsv'"
-
-# COPY 99997497
-# Time: 1320024.670 ms (22:00.025)
+# load data
```
rschu1ze, line 5: For this to work, the postgres client needs to be installed. Please grep `apt install` in the ClickBench repository for other databases which install stuff.

index.html
```diff
@@ -432,7 +432,7 @@
...
...],[0.245723,0.038173,0.037376],[0.306797,0.042102,0.042718],[0.418011,0.242568,0.23039],[0.551909,0.282084,0.278836],[0.180043,0.02662,0.025371],[0.165209,0.024473,0.023097],[0.453139,0.284364,0.283832],[0.535552,0.393579,0.39756],[0.30116,0.092113,0.091926],[0.257973,0.100089,0.098468],[0.398887,0.26349,0.253034],[0.578082,0.426437,0.425838],[0.434567,0.263478,0.266679],[0.379118,0.262621,0.264961],[0.729468,0.564334,0.562802],[0.602923,0.464127,0.465615],[1.33815,0.996799,0.993801],[0.173591,0.013425,0.013638],[1.68838,0.515917,0.516346],[0.437349,0.294343,0.296138],[1.99367,0.417726,0.417281],[8.64109,1.77034,1.77321],[0.264145,0.104584,0.105626],[0.238236,0.095826,0.095679],[0.287271,0.112243,0.108904],[0.755408,0.612582,0.61124],[6.75692,6.60809,6.60236],[0.23476,0.07472,0.067433],[0.382783,0.239876,0.239216],[0.438111,0.296216,0.288612],[1.5727,1.69792,1.40727],[1.55361,1.28201,1.28065],[1.49317,1.40405,1.38793],[0.579766,0.41326,0.413215],[0.226025,0.062452,0.059734],[0.195866,0.024387,0.041084],[0.198459,0.025349,0.026228],[0.323418,0.091749,0.088929],[0.172858,0.019531,0.019639],[0.172619,0.019008,0.022273],[0.194027,0.027197,0.024702]],"source":"hydra/results/hydra.json"}
```
rschu1ze, line 435: No need to update index.html by hand, it is automatically updated by a CI job based on *.json files in `results/` subfolders.

wuputah: got it, I'll undo this change.

hydra/benchmark.sh
wuputah: hmm, ok it seems the pattern is to leave a commented-out command for these.

wuputah: updated.

wuputah:
You can run it locally get a connection string using `hydra config`, but the benchmark used a paid cloud instance and was tested from the same cloud to eliminate effects of latency. A connection is available on all paid instances.

rschu1ze:
One of the goals of ClickBench is reproducibility. If I read correctly, the trial token isn't sufficient to run the benchmark, one needs a paid instance. Fair enough.

So if I subscribe, it will show the connection string somewhere?

> You can run it locally and get a connection string using hydra config

I didn't understand if that applies to the free or the paid version. I also didn't understand what "hydra config" is, where to get it and how to run it.

wuputah:
> So if I subscribe, it will show the connection string somewhere?

Yes, the connection string is displayed quite prominently on each project's dashboard, e.g.:

image
 
> I didn't understand if that applies to the free or the paid version. I also didn't understand what "hydra config" is, where to get it and how to run it.

The `pip install hydra-cli` provides a local `hydra` command which will allow you to run things locally with the free trial. Running the benchmark isn't the intention here and `hydra config` isn't really in the happy path, but is there for those that need it.

rschu1ze:
Regarding the free version:

I installed `hydra-cli`, went through `hydra setup`, then `hydra config` to obtain the connection string but then hit

```
/data/ClickBench2/hydra (update-hydra =) $ DATABASE_URL="postgresql://postgres:esyD31pH@localhost:52488/postgres" ./benchmark.sh
psql: error: connection to server at "localhost" (127.0.0.1), port 52488 failed: Connection refused
        Is the server running on that host and accepting TCP/IP connections?
```

It obviously connects to localhost but I couldn't figure the remote URL out. I think that is what you meant that it won't work with the free version.

I then tried to register for a paid version. I signed out and in again with my other mail address (my work mail address from Google didn't work, see above). I then see this screen:

grafik

I then click "Register Cloud Access" and it states that it sent a mail with a confirmation code. This is the second time I am asked for a confirmation code (the first time, I entered it properly). I am waiting since 20 minutes now but didn't receive another code. The mail address is `robert.hydra@mailbox.org` in case you like to check.

wuputah:
You'll need to run `hydra start` to have the local instance Postgres running.

rschu1ze:
Maybe it helps if the README.md file could specify the steps to reproduce the benchmark at more detail. Basically write it for someone who has never used Hydra.

wuputah:
Apologies, all of this is quite nascent so there appears to be a bug here with the local connection, I'll try to get a fix out for that ASAP.

The test itself was done on a paid instance where the credential would be readily available. Right now we are manually onboarding customers, so the product is not available without talking to us, which is what "Request Cloud Access" should have started. Happy to get you access if you wish to try it.

wuputah:
OK, that bug has been fixed. You can update with `pip install --upgrade hydra-cli` or as desired for your python toolchain.

I'll update the README with more instructions, both for local and cloud.

hydra/README.md
```diff
@@ -1,16 +1,59 @@
...
+3. Install Docker.
+4. Sign up for Hydra at https://start.hydra.so/get-started and copy the token to your clipboard
+5. Run `hydra setup` and paste in your token
+6. Run `hydra start` to start the local instance
+7. Run `hydra config` to obtain the connection string and set it `DATABASE_URL` in your local environment,
+   or pass it in when you run the benchmark.
+8. Run the benchmark.
+
+    ```
+    DATABASE_URL="..." ./benchmark.sh
```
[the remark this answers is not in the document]
wuputah: This seems like the same bug with the old CLI version, can you check `hydra --version`? If you have updated on the same machine, you'll then need to `hydra restart`, which should resolve the issue.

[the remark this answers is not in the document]
rschu1ze: I am fine with submitting the results, let's just make sure that the actually used compute option is specified in hydra.json (please see my other comment).

hydra/results/hydra.json
```diff
@@ -0,0 +1,58 @@
+{
+    "system": "Hydra",
+    "date": "2025-03-04",
+    "machine": "cloud",
```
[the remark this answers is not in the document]
wuputah: all set

PR #326, opened 2025-03-10
Update Postgres to PG17 and configure basic settings for unindexed PG

JelteF:
I was looking into Postgres performance on ClickBench and realized a bunch of things that could be improved. The primary changes I made were:

1. Upgrade to PG17
2. Configure the basic CPU & RAM related settings also for the untuned Postgres. Postgres defaults are chosen for extremely low resource usage, not for production systems.
3. Configure the settings from 2 based on available memory and CPU, instead of hardcoding. In a follow up PR I plan to submit numbers for c6a.metal.
4. Rename "Postgres (tuned)" to "Postgres (indexed)", because now both configurations have some basic tuning.

P.S. The numbers from the non-indeed Postgres were totally out of date, they were not rerun since the indexes were moved to a separate benchmark. This change makes the match reality much closer.


postgresql/benchmark.sh
```diff
@@ -1,35 +1,51 @@
...
 
+set -eux
+
+PGVERSION=17
+
 sudo apt-get update
-sudo apt-get install -y postgresql-common
-sudo apt-get install -y postgresql-14
+sudo apt install -y postgresql-common
+sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
```
rschu1ze, line 9: Could you please add "https://wiki.postgresql.org/wiki/Apt" as a comment behind l. 9?

postgresql/benchmark.sh
```diff
@@ -1,35 +1,51 @@
 #!/bin/bash
 
+set -eux
+
+PGVERSION=17
+
 sudo apt-get update
-sudo apt-get install -y postgresql-common
-sudo apt-get install -y postgresql-14
+sudo apt install -y postgresql-common
```
rschu1ze, line 8: Perhaps we should un-install existing Postgres first?

postgresql/benchmark.sh
```diff
@@ -1,35 +1,51 @@
...
+sudo /usr/share/postgresql-common/pgdg/apt.postgresql.org.sh
+
+sudo apt-get update
+sudo apt-get upgrade
+sudo apt-get install -y postgresql-$PGVERSION
+
+memory=$(awk '/MemTotal/ {print $2}' /proc/meminfo)
```
rschu1ze, line 15: For my understanding: Are the parameters in l. 15-20 best practices / commonplace? Perhaps add a comment to some reference (Stackoverflow?) which recommends them?

postgresql/run.sh
```diff
@@ -6,8 +6,9 @@ cat queries.sql | while read -r query; do
     sync
     echo 3 | sudo tee /proc/sys/vm/drop_caches
 
-    echo "$query";
-    for i in $(seq 1 $TRIES); do
-        sudo -u postgres psql test -t -c '\timing' -c "$query" | grep 'Time'
-    done;
-done;
+    echo "$query"
```
rschu1ze, line 9: I am fine with the change in this file, however the previous state looked more like Python / C++. I guess we are now using some bash idioms.

postgresql-indexed/results/c6a.4xlarge.json
```diff
@@ -0,0 +1,58 @@
+{
+    "system": "PostgreSQL (indexed)",
```
rschu1ze, line 2: "indexed" --> "with indexes"?

postgresql/run.sh
JelteF: Yeah, this change is mainly to reuse the same connection for the hot run. I did a similar trick for pg_duckdb.

postgresql/benchmark.sh
JelteF: Lot's of people have their own exact preferences, and the best ofcourse depends on the workload. But it's commonly agreed upon that the defaults are terrible. I didn't spend much time on finding the best, I mainly kept them close to whatever we already had in the "tuned" postgres. I added some comments about these settings and some links now.

postgresql/benchmark.sh
JelteF: Done

postgresql/benchmark.sh
JelteF: I don't think this necessary, you can install multiple PG versions at the same time. `postgresql-common` is primarily needed to get this install script.

PR #327, opened 2025-03-10
Update MotherDuck results to reflect current offering

hrl20:
MotherDuck recently introduced 3 instance types: pulse, standard, and jumbo (details: https://motherduck.com/blog/introducing-motherduck-for-business-analytics/).

So updating to show results for all 3 instance types, and removing the previous result which is outdated, since it's based on an instance type/configuration that is no longer offered. 

Results are obtained by running from a VM in aws us-east-1.


index.html
```diff
@@ -444,14 +444,7 @@
...
...6700056489,0.12905511599819874,0.11581274600030156],[0.06644965200030128,0.10702265000145417,0.06476570199811249],[0.11048688200025936,0.07675166399712907,0.07593394799914677],[0.2186288560005778,0.2090998989988293,0.18735763899894664],[1.3868503670018981,1.3703306460010936,1.2348434389969043],[0.7405084229976637,0.7247317809997185,0.729713893000735],[0.18527704200096196,0.1730873589986004,0.15151636199880159],[0.20919774299909477,0.191980165996938,0.1933938189977198],[0.7669437040021876,0.7439510249969317,0.7487478670009295],[0.8591277329978766,0.8134900770019158,0.8157381170021836],[0.8418528250003874,0.8204738799977349,0.8599285560012504],[0.2758991349983262,0.2533942770023714,0.24527843900068547],[0.05365405100019416,0.045813198001269484,0.04610075799791957],[0.03199732199936989,0.022308995001367293,0.021592714001599234],[0.030769754001084948,0.030530198997439584,0.03006599299988011],[0.09920943899851409,0.12626113100122893,0.08831893900060095],[0.023186009002529318,0.015893660998699488,0.013306151999131544],[0.03293053500237875,0.012570572002005065,0.012848910002503544],[0.025673778996861074,0.015225869999994757,0.016331293001712766]],"source":"motherduck/results/result.json"}
```
rschu1ze, line 447: index.html will be auto-generated from the json files in the "results/" subfolders. Please revert the changes in index.html and instead add / modify the existing files in "motherduck/results/" (there needs to be one file per tier).

hrl20: I've removed the accidental index.html update- This pr only changes the result files.

PR #331, opened 2025-03-11
Add results for pgpro_tam Postgres extension

InnerLife0:
pgpro_tam is an extensible table access method for vanilla Postgres

It supports multiple data formats, storage managers and execution engines.
More: https://hub.docker.com/r/innerlife/pgpro_tam

pgpro_tam extension allows to use different pluggable modules.

Data format modules used in tests:
1) Parquet
2) Feather (Apache Arrow file-based IPC)

Storage manager modules used in tests:
1) Local storage
2) Local storage with in-memory file mirroring

Execution engine modules used in tests:
1) DuckDB

For parquet the default compression is ZSTD3.
One of tests uses 'parallel' for loading. In case of inserting a data from one big file, the speed is far from maximum, but a deal is a deal. We do not split the file because it is not recommended by rules. The same test uses higher block size.
We also do not use fine tuning as it is not recommended. Tuning each column with own lightweight encoding would give better results.
Feather format is presented to share our experience. Using an uncompressed format gives gains on some queries, but on average it is worse than parquet.
Since AWS EC2 is generally unavailable in our region, the tests were run on similar virtual machines. For example, on "16 vCPU 32GB" (AMD EPYC 9354) instead of "c6a.4xlarge".


pgpro_tam/README.md
```diff
@@ -0,0 +1,22 @@
+###pgpro_tam is an extensible table access method for vanilla Postgres
```
rschu1ze, line 1: Minor: let's remove the leading # ... that's for headlines

InnerLife0: Fixed

pgpro_tam/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+It supports multiple data formats, storage managers and execution engines.
+More: https://hub.docker.com/r/innerlife/pgpro_tam
```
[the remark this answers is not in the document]
InnerLife0: This is a new project and the decision on whether to release the code to open source has not yet been made.

pgpro_tam/results/30vCPU.feather_fmt.mem_fd_smgr.json
```diff
@@ -0,0 +1,58 @@
+{
```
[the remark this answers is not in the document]
InnerLife0: Understand.
Answering to this question and question about c6a.4xlarge.
Will it be OK to use some configuration which already exist in tests? '16 vCPU 128GB', for example. I can update readme with hardware specs behind virtual machine, if needed. Another option is to use 32GB RAM which is equal to c6a.4xlarge.
We still want to keep results for smaller machine, but AWS ec2 is generally unavailable in our region and it was not such easy to get access to c6a.metal presented here.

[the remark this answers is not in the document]
rschu1ze: I see. Yes, usage of equivalent configurations is absolutely okay. Perhaps add a remark in the README so that future readers will know.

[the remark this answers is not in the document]
InnerLife0: I have updated the test results and readme. Please review if everything is ok.

pgpro_tam/benchmark.sh
```diff
@@ -0,0 +1,67 @@
+#!/bin/bash
+
+#Usage:
+#./benchmark.sh parquet_fd
```
rschu1ze, line 4: The invocation is is too confusing what is being measured here. What does "fd" mean? I first thought it means "feather" but there is a separate feather invocation (l. 7). If "fd" occurs in all four invocations (l. 4-7), why do we need to specify it at all?

pgpro_tam/results/16vCPU.parquet_fmt.fd_smgr.json
```diff
@@ -0,0 +1,58 @@
...
+    "cluster_size": 1,
+    "comment": "",
+
+    "tags": ["C", "column-oriented", "PostgreSQL compatible"],
+
+    "load_time": 896,
+    "data_size": 9690683705,
+
+    "result": [
+        [0.121965,0.055939,0.050485],
```
rschu1ze, line 14: My measurements on a _c6a.4xlarge` with gp2 disk.

```
0.321687,0.046211,0.04321],
0.344998,0.069772,0.068691],
0.391883,0.112431,0.113001],
0.416684,0.100819,0.099516],
0.777907,0.506035,0.512778],
1.09328,0.792829,0.818029],
0.32257,0.060854,0.061807],
0.346361,0.076369,0.071959],
0.943637,0.632592,0.623626],
1.11601,0.793496,0.799572],
0.543303,0.22557,0.223995],
0.597173,0.263162,0.266207],
1.10562,0.796811,0.803204],
1.59857,1.30935,1.21759],
1.1899,0.878421,0.882841],
0.888102,0.574016,0.573218],
1.88187,1.52702,1.53991],
1.62617,1.29211,1.28645],
3.63748,2.63182,2.67271],
0.384733,0.105581,0.103658],
5.0365,2.09136,2.04135],
5.6607,1.30087,1.30787],
10.9207,2.35458,2.2812],
31.7888,10.4792,10.2795],
0.552504,0.262987,0.268879],
0.696989,0.434071,0.426577],
0.548206,0.271086,0.273945],
4.85798,2.05637,2.08517],
11.1605,10.7678,10.5724],
0.342533,0.090584,0.093431],
1.33926,0.91594,0.925731],
4.6345,1.04401,1.01668],
5.30164,2.81433,2.84594],
5.83912,3.7593,3.7557],
5.7988,3.77693,3.78104],
1.10924,0.829021,0.830779],
0.523506,0.235344,0.23369],
0.562088,0.319885,0.303772],
0.392827,0.12452,0.104705],
0.652588,0.36169,0.350525],
0.321632,0.055244,0.053685],
0.31652,0.065746,0.053403],
0.34475,0.075701,0.058059],
```

If you like, I can run the measurements (all four benchmark variants) on my local c6a.4xlarge instance and copy-paste results. That would increase compareability with other benchmarks. Let me know what you think - I'd be fine either way.

pgpro_tam/benchmark.sh
InnerLife0: pgpro_tam works with multiple formats and storage managers. 'fd' states for ppg_fd_smgr and points to classic 'fd.c' from postgres, which implements basic operations with storage. There is another storage manager - ppg_mem_fd_smgr, which extends possibilities of the previous one. Keeping only 'mem' for this may cause misunderstanding, as it's not a stateless in-memory storage.
If it makes sense, they can be renamed, but anyway, results and readme already contain a human-readable description.

pgpro_tam/results/16vCPU.parquet_fmt.fd_smgr.json
InnerLife0: ~~Damn, that disks are too slow, so I don't know from where I can dig the same.~~

If it's OK to leave current results as is, then it's a better option. Despite the fact you can kindly help us with updating our results on every PR, we need somehow to now, are our new results better than current results on board.
If current results break the rules and recognized as incomparable, then asking you is the only good option we have.
Thanks.

rschu1ze:
> @rschu1ze I'm curious to know if it will be possible to replace the c6a.metal results in the future with the results from the machine used [here](https://github.com/ClickHouse/ClickBench/blob/main/crunchy-bridge-for-analytics/results/crunchy-bridge-analytics-256.json?rgh-link-date=2025-03-12T19%3A04%3A59Z).
> Since it is difficult for us to provide updated results on c6a.metal and since we don't want to clutter the board with our results, I just want to remove c6a.metal and provide the result on a similar machine (not now, but in the future pull requests).

Yes, doing so in a future PR would definitely be nice!

PR #316, opened 2025-03-04
CrateDB: Update results

hammerhead:
Results for CrateDB are almost three years old, so here is an update.

Summary of the changes:
1. Update installation method to DEB package: This gives greater flexibility to test different versions, such as preview releases.
2. Fix query 29: Extracting the referrer domain using `REGEXP_REPLACE` didn't capture the correct group, which needs to be referred to as `$1`.
3. Added a tuned version: It uses a higher number of shards and stores empty strings as `NULL`. `SELECT COUNT(DISTINCT x)` is rewritten as `SELECT COUNT(*) FROM (SELECT x FROM t GROUP BY 1)` which is more performant in CrateDB.

PR #345, opened 2025-04-08
CrateDB: Fix naming for tuned results

hammerhead:
This is a small fix following #316 where the tuned version wasn't named differently from the standard results.


cratedb/results/c6a.4xlarge-tuned.json
```diff
@@ -1,5 +1,5 @@
 {
-    "system": "CrateDB",
+    "system": "CrateDB (tuned)",
```
rschu1ze, line 2: Easiest PR I reviewed this week :-)

PR #352, opened 2025-04-17
Fix sed aliasing in `generate-results.sh` for macOS

AdamGS:
By default, aliases are not expanded in bash scripts, meaning the macos `sed` is still being used in the rest of the script even if `gsed` is available.
This change allows aliases to be expanded for the purposes of running this script, verified this behavior and fix on my macbook.


AdamGS:
@rschu1ze thanks for the quick tunaround and thank you for maintaining this project!

PR #356, opened 2025-05-01
fix generate-results.sh so that it can be run on a mac

jinxidoru:
This is a pretty small change to the generate-results script that makes it so it'll run on a Mac. The sed command there is slightly different and doesn't like range delete command for some reason. I verified that this new version of the command will also work on gnu sed.

### DIFF
--- a/generate-results.sh
+++ b/generate-results.sh
@@ -25,7 +25,7 @@ fi
     done
 
     echo ']; // end of data'
-    sed '0,/^\]; \/\/ end of data$/d' index.html
+    sed -n '/^\]; \/\/ end of data$/,$p' index.html | sed '1d'
 
 ) > index.html.new
 


rschu1ze:
@jinxidoru Would you sign the CLA first? Thanks.

jinxidoru:
I signed it when I first posted the PR but it does not seem to register the fact that I signed it. Not sure what to do at this point.

jinxidoru:
I signed it when I first posted the PR but it does not seem to register the fact that I signed it. Not sure what to do at this point.

jinxidoru:
I just tried it again. It had me log in to GitHub then told me that I'd already signed it while redirecting me to GitHub.

rschu1ze:
I tried locally, it looks like we are now generating something different on Linux than before.

jinxidoru:
@rschu1ze - Are you sure about that? I just ran it to double check and got identical results. The most recent benchmark commit didn't update the index.html, just some of the results files. So I'm wondering if perhaps what you're seeing is simply the update of the index.html based upon the recent results changes, because I'm not seeing at all how my change could be giving different results on linux.

rschu1ze:
It sounds more appealing to force users to install GNU sed on macOS than having to maintain a lowest-common sed pattern for both versions. I did that in https://github.com/ClickHouse/ClickBench/pull/360.

PR #357, opened 2025-05-03
Update Parquet results for Hyper

caetanosauer:
This fixes an issue with how we were handling text columns that are stored as byte arrays in Parquet. With a new option `binary_as_text`, these columns are correctly interpreted as text rather than binary, and the query results are now correct.

Once again kudos to Mark Raasveldt (@Mytherin) who reported this months ago.


rschu1ze (approved):
I was able to reproduce the results locally, thanks!

PR #375, opened 2025-05-21
Add duckdb using vortex files

joseph-isaacs:
Add duckdb run using vortex as the file format.


joseph-isaacs:
Updated @rschu1ze Thanks for reviewing :)

PR #413, opened 2025-06-23
Add Firebolt

hall-alex:
Add the Firebolt database system: https://www.firebolt.io


moshap-firebolt:
Hi @rschu1ze - I am CTO of Firebolt, and I confirm that @hall-alex is also officially affiliated with Firebolt. Hereby the restriction you pointed out to does not apply, as submission has been done by Firebolt itself. But to make it even more explicit - you can consider this comment as a written consent for anyone to reproduce the Clickbench runs in this submission and to verify the numbers.

firebolt/benchmark.sh
```diff
@@ -0,0 +1,42 @@
+#!/bin/bash
+
+# Download the hits.parquet file
+echo "Downloading dataset..."
+rm -rf data
+mkdir -p data
+wget -P data "https://datasets.clickhouse.com/hits_compatible/hits.parquet"
```
rschu1ze: Could you add `--continue` for consistency with other benchmark scripts?

firebolt/results/c6a.metal.json
```diff
@@ -0,0 +1,54 @@
+{
+    "system": "Firebolt",
```
rschu1ze, line 2: My measurements:

```
    "load_time": 156.778640,
    "data_size": 15716159954,
    "result": [
        [0.004784,0.001658,0.002317],
        [0.088994,0.013734,0.015525],
        [0.09544,0.015088,0.014866],
        [0.142752,0.013368,0.015307],
        [0.048307,0.04115,0.042862],
        [0.689594,0.042683,0.041577],
        [0.046456,0.014687,0.013073],
        [0.017039,0.01301,0.014065],
        [0.473524,0.046856,0.047489],
        [1.150012,0.067616,0.06661],
        [0.487908,0.024363,0.022572],
        [0.683946,0.024926,0.025723],
        [0.064943,0.064546,0.065673],
        [1.554241,0.121868,0.130354],
        [1.035027,0.071879,0.076728],
        [0.047935,0.044905,0.04532],
        [0.099997,0.091907,0.087903],
        [0.074742,0.073031,0.073879],
        [3.541261,0.149745,0.152064],
        [0.017149,0.010736,0.011055],
        [8.673245,0.051971,0.051208],
        [10.253905,0.045684,0.042594],
        [19.109284,0.080522,0.077431],
        [59.272084,0.123945,0.082282],
        [2.666837,0.04154,0.033858],
        [0.044239,0.028846,0.040567],
        [0.033288,0.027871,0.035536],
        [8.629192,0.092781,0.089545],
        [7.405405,0.242182,0.237926],
        [0.098412,0.047189,0.048665],
        [1.487216,0.052243,0.045211],
        [4.98714,0.060798,0.064355],
        [3.592003,0.250428,0.259408],
        [0.362568,0.286633,0.277147],
        [0.289864,0.290742,0.315497],
        [0.118459,0.049228,0.052029],
        [8.877934,0.02289,0.021489],
        [7.793583,0.017102,0.01522],
        [9.192633,0.015729,0.013637],
        [17.781422,0.031324,0.030222],
        [2.912249,0.015003,0.018048],
        [2.543615,0.015537,0.015928],
        [2.259005,0.014457,0.016211]
```

Numbers are very close to your measurements 👍

firebolt/benchmark.sh
hall-alex: I see you already added it, thanks 🙏.

rschu1ze:
Added measurements on c6a.4xlarge which is what lots of databases in Clickbench use as underlying hardware.

hall-alex:
Thanks @rschu1ze for the ultra-fast turn-around and for adding numbers for c6a.4xlarge 🙏.

moshap-firebolt:
Hi @rschu1ze - just to follow up on the [terms](https://www.firebolt.io/trial-subscription-agreement) you cited yesterday - the DeWitt clause was there by mistake - we removed it years ago, but somehow missed it getting back during one of updates. It's all taken care of now - this clause is gone.

firebolt/run.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+# Disable the result and subresult caches. Enable the scan-cache.
+QUERY_PARAMS="enable_result_cache=false&enable_subresult_cache=false&enable_scan_cache=true&output_format=JSON_Compact"
```
chhetripradeep, line 4: @hall-alex since i can't find this in the firebolt documentation, can you elaborate what does `enable_scan_cache=true` does ?

rschu1ze: I think it's [this cache](https://docs.firebolt.io/reference/system-settings#scan-cache) (off by default). Other databases utilize similar caches, e.g. ClickHouse has the "Uncompressed Cache" (docs [1](https://clickhouse.com/docs/operations/caches), [2](https://presentations.clickhouse.com/meetup60/overview-of-caching.pdf), also off by default).

I would say this kind of caches falls under ClickBench [rule](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#caching) `It is okay if the system performs caching for source data (buffer pools and similar). If the cache or buffer pools can be flushed, they should be flushed before the first run of every query.`. It is a grey zone though ... the thing about the scan cache (Firebolt) / Uncompressed Cache (ClickHouse) is that they essentially nullify the effect of flushing the OS page cache (`echo 3 | sudo tee /proc/sys/vm/drop_caches > /dev/null`) which happens between query runs.

For the sake of better comparability, it would be a bit fairer if there was
- a base Firebolt ClickBench configuration (where the database run with its default settings as far as that is possible), and
- a tuned Firebolt ClickBench configuration (for example with additional caches and physical tuning).

Corresponding default and tuned configurations exist for the ClickHouse submission in ClickBench (however with the Uncompressed Cache disabled in both).

hall-alex: @rschu1ze, yes that's how we interpreted the [rule](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#caching) you cite.

We can of course follow up with a submission as you suggest -- `c6a.4xlarge` also needs some investigation. I'm travelling currently, but should get to it next week.

@moshap-firebolt, @wagjamin, @leonard-firebolt, as FYI.

moshap-firebolt: Thanks @rschu1ze and @chhetripradeep - we did pay close attention to the rules of Clickbench, and made our best effort to be compliant with all the rules. There are two particular sections of the rules - [Installation and Fine Tuning](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#installation-and-fine-tuning) and [Caching](https://github.com/ClickHouse/ClickBench?tab=readme-ov-file#caching) which are in a bit of contradiction. The installation asks `It's better to use the default settings`, while caching gives instructions to disable result and subresult caches, but gives ok to caching of source data (which is what [enable_scan_cache](https://docs.firebolt.io/reference/system-settings#scan-cache) does).

Following @rschu1ze suggestion to remove the ambiguity and have better comparability - I think we should rename current submission to `Firebolt (scan cache)` and then submit another entry for just `Firebolt` which will still have to have result and subresult caches disabled.

Does that sound like a good plan ? If so - I will do a PR with rename now.

moshap-firebolt: I also want to add - that we will follow up with thorough documentation on how scan cache works in Firebolt (beyond the paragraph in the system setting page), to clarify what exactly it does, and how it is really equivalent to buffer pool functionality.

rschu1ze: > Following @rschu1ze suggestion to remove the ambiguity and have better comparability - I think we should rename current submission to Firebolt (scan cache) and then submit another entry for just Firebolt which will still have to have result and subresult caches disabled.

Yes, sounds good, thanks. My main point (and I probably should have phrased it better) was that all submissions are ideally benchmarked with their default setting (i.e. `It's better to use the default settings` from the installation section). 

But I agree that the ClickBench rules leave some wiggle room. We have been discussing a few changes to ClickBench internally. Ideas include:

- simpler provisioning the benchmark scripts, probably with automatic re-runs (to lower the review and maintenance burden),
- adding a rule which requires that the database is shutdown or killed between queries (besides clearing the OS page cache) to ensure cold runs are really cold. It probably makes sense to continue measuring the current practice where only the OS page cache is cleared as "warm" runs, in distinction to true "hot" runs.
- many databases are so fast that runtime differences between databases blur. We should probably add more queries, make existing queries more complex (while keeping the benchmark characteristics, i.e., aggregation heavy), or increase the dataset size.

Anyways, there would be a community discussion first, and of course all changes would be applied to _all_ submissions.

PR #421, opened 2025-06-30
Fix result rendering and correct load_time / data_size for TCHouse-C

amosbird:
Add `tuned: no` or else the recently submitted result cannot be rendered correctly. 

Also the load_time and data_size is corrected. Ref: https://github.com/ClickHouse/ClickBench/pull/411/commits/c02e64c66ad9b6448119f6d35cf9cc17acd5becd


amosbird:
@rschu1ze Would appreciate your review and merge, thanks!

PR #579, opened 2025-08-26
Update to Sail 0.3.3

shehabgamin:
Updated to Sail 0.3.3 and also added sail-partitioned.

Didn't update any results, wasn't sure if I'm supposed to do that or not.


sail-partitioned/queries.sql
rschu1ze, line 1: The files in sail-partitioned/ mostly duplicate the files in sail/.
May I ask you to create symbolic links where possible?

sail-partitioned/benchmark.sh
```diff
@@ -0,0 +1,63 @@
+#!/bin/bash
+
+# Install
+
+export DEBIAN_FRONTEND=noninteractive
+
+echo "Set Timezone"
```
rschu1ze, line 7: Why is it needed to set a time zone?

sail-partitioned/benchmark.sh
```diff
@@ -0,0 +1,63 @@
...
+export TZ=Etc/UTC
+sudo ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
+
+echo "Install Rust"
```
rschu1ze, line 11: It would be nice if we could either use the Rust packages that ship with Ubuntu 24.10 (if not too old), or install at least in a local directory (not the user home directory).

sail-partitioned/benchmark.sh
shehabgamin: When you run Sail on Amazon Linux, you may encounter the following error:

> `failed to get system time zone: No such file or directory (os error 2)`

The reason is that `/etc/localtime` is supposed to be a symlink when retrieving the system time zone, but on Amazon Linux it is a regular file.

There is a [GitHub issue](https://github.com/amazonlinux/amazon-linux-2023/issues/526) for this problem, but it has not been resolved yet.

sail-partitioned/benchmark.sh
shehabgamin: @rschu1ze I just copied what I saw in DataFusion:
https://github.com/ClickHouse/ClickBench/blob/main/datafusion/benchmark.sh#L3-L7

Also, It looks like modifying the user home directory was a recent addition in https://github.com/ClickHouse/ClickBench/pull/496
```
export HOME=${HOME:=~}
```

sail-partitioned/queries.sql
shehabgamin: Do you have an example to share?

Looking at `clickhouse-parquet-partitioned`, it's also a duplicate of the files in `clickhouse-parquet`:
- https://github.com/ClickHouse/ClickBench/tree/main/clickhouse-parquet-partitioned
- https://github.com/ClickHouse/ClickBench/tree/main/clickhouse-parquet

alexey-milovidov: @rschu1ze, this does not matter, and we didn't use symlinks before. Copy-paste is ok.

sail-partitioned/benchmark.sh
alexey-milovidov: @rschu1ze, this does not matter, ClickBench runs in short-lived VMs.

alexey-milovidov (approved):

Issue #517, opened 2025-07-06
Add Spark accelerators


Iskander14yo:
### Motivation
There are several Spark plugins and extensions that claim to offer significantly better performance than vanilla Spark for analytical workloads -for example, **Apache Comet** (based on DataFusion), **Blaze** (DataFusion), and **Apache Gluten** (mainly Velox, ClickHouse). Since ClickBench already allows GPU-based engines (#410), NVIDIA RAPIDS could also be considered.

I believe these are worth evaluating for a few reasons:
- **Independent benchmark**: These accelerators are typically benchmarked only on their own TPC-H/TPC-DS variants. Including them in ClickBench would offer a more independent comparison. This is valuable to both users and developers.
- **Distributed vs. single-node overhead**: Some of these tools have single-node counterparts (e.g., DataFusion). Measuring how much overhead comes from different factors (such as Spark’s architecture, quality of implementation) is useful in itself.
- **Spark user insights**: For Spark users, this shows how much they can gain (or lose) by adopting such plugins (which are mostly easy to install having already deployed Spark) to compare with other distributed engines and Spark itself.

### Request
Would you be open to including these in ClickBench? If I checked correctly, no similar issue has been created yet.
If yes, I’d be glad to **take on** this task and submit the necessary PRs incrementally (likely one per engine).

Iskander14yo:
Can I ask some questions?

Do we pursue a goal to put these engines in the most equal conditions, or is that not necessary? I mean mainly **Spark versions** here — since hardware is fixed, and I haven’t identified other major factors that could impact results yet.

ClickBench isn’t Spark-specific, so it seems fine to use whatever works best per engine. But if we want comparisons to be more fair (and helpful for Spark users), maybe it makes sense to _align_ Spark versions across vanilla Spark, Comet, Gluten, etc.

So first option is to keep Spark version _consistent_ within the Spark “family,” and allow exceptions with clear notes.
Another option is to use _"latest stable Spark version per engine"_ approach. My thoughts were:
- Current Spark version in repo is 4.0 and now it is supported only by Comet and only for testing;
- Version alignment might be hard to maintain (e.g., today Gluten-on-Clikchouse only supports Spark 3.2/3.3);
- Patch versions also differ often, though it doesn't seem critical.

Let me know which direction seems reasonable for you. Currently I'm leaning towards second option because it's easier to maintain and still "fair" (we use same VM anyway and let each engine use latest Spark version it supports).

[...]

PR #614, opened 2025-09-08
add Apache Auron

Iskander14yo:
This PR adds [Apache Auron](https://auron.apache.org/) (formerly Blaze) as a backend for Spark in analytical workloads. Base issue - https://github.com/ClickHouse/ClickBench/issues/517.

Structure is similar to Apache Comet (`spark-comet/`) with minor differences.

This PR is marked as Draft since there are no results (I can't run bench currently) but scripts are already tested for Ubuntu 24.04.

### DIFF
--- a/spark-auron/template.json
+++ b/spark-auron/template.json
@@ -8,5 +8,4 @@
       "column-oriented",
       "Spark derivative"
     ]
-  }
-  
+}


rschu1ze (commented):
@Iskander14yo Would you merge latest `main` into your PR and create a template.json file (like now exist in all other entries).

Iskander14yo:
Updated branch and added template.json (copied from Comet)

spark-auron/query.py
```diff
@@ -0,0 +1,66 @@
...
+import pyspark.sql.functions as F
+
+import psutil
+import re
+import sys
+import timeit
+
+
+query = sys.stdin.read()
+# Replace \1 to $1 because spark recognizes only this pattern style (in query 28)
```
alexey-milovidov, line 21: Replace it in queries.sql

spark-auron/query.py
```diff
@@ -0,0 +1,66 @@
...
+query = re.sub(r"""(REGEXP_REPLACE\(.*?,\s*('[^']*')\s*,\s*)('1')""", r"\1'$1'", query)
+print(query)
+
+# Calculate available memory to configurate SparkSession
+ram = int(round(psutil.virtual_memory().available / (1024 ** 3) * 0.7))
```
alexey-milovidov, line 26: Does not work on machines with 1 GB of RAM or less.

alexey-milovidov:
It is 99% ok, only two small changes needed...

spark-auron/query.py
Iskander14yo: Initially, I copied this approach from vanilla Spark setup. At the same time, I like it - this way we can keep original SQL unchanged and localize Spark-specific handling in `query.py` for traceability.
If the project convention is to adjust SQL, I can do that (and for other Spark engines too). Ready to follow the preferred approach.

spark-auron/query.py
Iskander14yo: This is similar to previous comment. I matched the vanilla Spark memory policy to keep conditions consistent and results comparable across Spark engines.
We may change this policy. In this case, we probably should do that for all Spark engines and re-run them for apples-to-apples numbers.

alexey-milovidov:
It's about the log size, make it less than 1 MB.

alexey-milovidov:
> Replace it in queries.sql

It is better, because we can easily see how queries differ between systems.

Iskander14yo:
OK, I'll try to get back to that in a couple of weeks.

PR #630, opened 2025-09-26
feat: introduce starrocks parquet

Thearas:
Introduce Starrocks parquet test.

I use some `CAST` at the projections of create-view statment in `create.sql`, since the query will throw error `Getting analyzing error. Detail message: Type (nested) percentile/hll/bitmap/json not support group-by.` without `CAST`.

It may slow down the result (? I don't know), let me know if we have other choice.


Thearas:
> Why closing?

Because the test method is not good (may slowdown the result):

> I use some CAST at the projections of create-view statment in create.sql, since the queries will throw error Getting analyzing error. Detail message: Type (nested) percentile/hll/bitmap/json not support group-by. without CAST.

PR #632, opened 2025-09-30
chore: update to sail 0.3.7

shehabgamin:
Also added empty results in `sail-partitioned/results`, because for some reason it's not being populdated


alexey-milovidov:
I confirmed with @rschu1ze that he will review this PR as soon as possible.

PR #633, opened 2025-10-03
Update QuestDB results to v9.1

puzpuzpuz:
Also changes type of IP columns to IPv4.


puzpuzpuz:
> I ran the script locally for `c6a.metal` and `c6a.4xlarge`. The former succeeded and I updated your measurements.

TBH I don't see any noticeable difference between your result and what was in the branch, but that's fine. At least, this means that I made no mistake when benchmarking.

> The latter somehow got stuck

This is puzzling. I've never seen the `waiting for rows to become readable...` loop to be stuck, but indeed it may take a while to finish (~1.5h total for the import on the smaller box). I'll try reproducing this.

PR #646, opened 2025-10-14
Add Apache Auron

Iskander14yo:
This MR closes (reapplies) #614.

@alexey-milovidov I made fixes as you requested. Eventually, both of them sound reasonable for me.

About log size: the current Auron release does not allow adjusting the logging level. While I probably can implement workarounds to reduce Auron’s log volume, I prefer to avoid that to prevent bias. Can we rerun the bench with Auron’s current logging?
If the run fails due to log volume we can probably wait for Auron 6.0. However, release date is unknown yet, which risks leaving Auron out of Clickbench indefinitely. I may have missed other options.

### DIFF
--- a/spark-auron/benchmark.sh
+++ b/spark-auron/benchmark.sh
@@ -31,7 +31,7 @@ wget --continue --progress=dot:giga $AURON_JAR_URL -O auron.jar
 
 # Run the queries
 
-./run.sh 2>&1 | tee log.txt
+./run.sh >log.txt 2>&1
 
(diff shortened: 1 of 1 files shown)


alexey-milovidov:
Piping the logs to /dev/null or to grep seems like an obvious solution, doesn't it?

alexey-milovidov:
@Iskander14yo, thank you, this is amazing! I will run it on all available machines...
Looking forward to merging the benchmarks of other Spark accelerators!

Issue #656, opened 2025-10-23
Cold run makes little sense for databases with in-memory and remote storage


puzpuzpuz:
The benchmark results feel more and more like a mess due to a number of issues.

1. As the title says, `Cold Run` results make no sense with in-memory and remote storage, so ideally they should not be shown on the generated html page.

2. The benchmark rules mention the following:
```
run.sh: a loop for running the queries; every query is run three times; if it's a database with local on-disk storage, the first query should be run after dropping the page cache;
```

This rule is not applied to some of the contestants that run on the local disk like [this one](https://github.com/ClickHouse/ClickBench/blob/318389533f6696852b0dccadb0fc5c10d9852a68/arc/run.sh) which produces wrong results in Combined and Cold Run categories.

3. The benchmark rules should specify the disk/volume type and its settings. The Machine field used to have `gp2 500GB` in the text - now it doesn't, so it looks like contestants can use any volume type, e.g. io2, making the Cold Run results questionable.

alexey-milovidov:
> As the title says, Cold Run results make no sense with in-memory and remote storage, so ideally they should not be shown on the generated html page.

Yes, it will make sense to hide in-memory systems from Cold and Combined results.

> This rule is not applied to some of the contestants that run on the local disk like [Arc](https://github.com/ClickHouse/ClickBench/blob/318389533f6696852b0dccadb0fc5c10d9852a68/arc/run.sh) which produces wrong results in Combined and Cold Run categories. For instance, Arc uses DuckDB as its query engine: it outperforms DuckDB+parquet in Cold Run by 3.08x, but it's slower than DuckDB in Hot Run by 2.7x - that's quite puzzling. More than that, many if the Arc's Cold Run query execution times are worse than the corresponding Hot Run times.

Did you find an error in Arc's benchmark scripts? They use a really big / complicated Python script, but at the first glance everything was alright. Let's figure it out.

> The benchmark rules should specify the disk/volume type and its settings. The machine field used to have gp2 500GB in the text - now it doesn't, so it looks like contestants can use any volume type, e.g. io2, making the Cold Run results questionable.

This is a wrong accusation. Here you see how the benchmark is run: https://github.com/ClickHouse/ClickBench/blob/main/run-benchmark.sh. The runs are automated to put every contestant in a fair position (for self-hosted entries). As far as I know, none of the systems used `io2` disks. No reason to share speculations.

> is it on purpose that the benchmark turned into цирк с конями?

[...]

PR #658, opened 2025-10-26
Prevent cheating in `Arc`

alexey-milovidov:
`Arc` results were incorrect. Highlighted by @puzpuzpuz in https://github.com/ClickHouse/ClickBench/issues/656


xe-nvdk:
Wow! you deleted the results for arc for this... we stated in our repo that we are in alpha technical preview and changes can happen... Cheez, this is a sign that we are doing something good. Instead of validating if something change in the code, you delete the results cause somebody say that we are cheating?

alexey-milovidov:
The reason is that the benchmark stopped reproducing. You can update the scripts to bring them back.

alexey-milovidov:
The scripts were also wrong - not compliant with the benchmark rules, as they didn't flush the filesystem cache between runs. This is qualified as cheating, as it makes the "cold" results artificially good.

xe-nvdk:
Ok, let me flush the os cache, set arc to disable query objects of duckdb and query cache to run the benchmarks again.

PR #661, opened 2025-10-26
Adding Arc results for c6a.4xlarge

xe-nvdk:
This is follow-up for https://github.com/ClickHouse/ClickBench/pull/658

Hey team. Nice sunday eh! Thank you to @alexey-milovidov to mark that the filesystem cache was enabled and that was originating not fair comparison. I want to clarify that wasn't on purpose or with the intention of cheat, and for that reason, we are posting the results for c6a.4xlarge again and **yes, cold run is triple that the original**, however, we are still happy cause we are perfoming betters than other systems in the category.

I fixed the script to point to the right endpoints, and also added a check for the storage, so, if you have aws-cli installed you can see what type of storage is running, also this is for eliminate any doubts regarding the storage used. 

Have a look, and let me know if you have any questions. 

Thank you

PR #669, opened 2025-10-27
Add more measurements for Arc

rschu1ze:
Added a few more measurements for Arc after the scripts [were re-submitted](https://github.com/ClickHouse/ClickBench/pull/661).


nwoolmer:
The deviation may be due to your use of integers in the filters instead of dates. I suspect if you use the same queries as [DuckDB (parquet, single)](https://github.com/ClickHouse/ClickBench/blob/main/duckdb-parquet/queries.sql), you should get the same results. 

Do you recall why you changed them before?

PR #685, opened 2025-11-10
Update results chDB, latest release 3.7.1

kafka1991:
Updated result sets for chDB 3.7.1


auxten:
The performance of chdb, chdb-parquet on c6a.metal seems getting worse?

auxten (approved):

Issue #667, opened 2025-10-27
Inconsistent use of database restarts across systems


lukasvogel:
I understand the motivation for restarting the database after every query. While I think considering the use of a buffer manager in a DBMS to be cheating is a bit strange, I agree that it makes a "cold" run less cold. 
If you want to require a restart before a "cold" run, that's fine by us. It's your benchmark, after all.

However, it’s not clear why this approach is applied selectively to only some databases but not others. For example, if I restart ClickHouse after every run, I get the following results (left: main branch; right: ClickHouse restarted before each query):

<img width="1994" height="1876" alt="Image" src="https://github.com/user-attachments/assets/bca3753c-3802-4c0c-9d10-e4f711c75a21" />

Most other databases that use a buffer pool (such as MySQL or Postgres) aren't restarted at all. It just seems inconsistent to only do this with the fast database systems.

Could you clarify the rationale behind which systems are restarted and whether this policy could be made consistent across all databases?

In case you're interested, here is the modified `run.sh` for ClickHouse:

```bash
#!/bin/bash
# Restart clickhouse so no cache can be used
sudo clickhouse stop  > /dev/null 2>&1

# Determine which set of files to use depending on the type of run
if [ "$1" != "" ] && [ "$1" != "tuned" ] && [ "$1" != "tuned-memory" ]; then
    echo "Error: command line argument must be one of {'', 'tuned', 'tuned-memory'}"
    exit 1
else if [ ! -z "$1" ]; then
    SUFFIX="-$1"
fi
fi

TRIES=3
QUERY_NUM=1
cat queries"$SUFFIX".sql | while read -r query; do
    [ -z "$FQDN" ] && sync
    [ -z "$FQDN" ] && echo 3 | sudo tee /proc/sys/vm/drop_caches >/dev/null

    sudo clickhouse start > /dev/null 2>&1
    for _ in {1..300}
    do
        clickhouse-client --query "SELECT 1"  > /dev/null 2>&1 && break
        sleep 1
    done

    echo -n "["
    for i in $(seq 1 $TRIES); do
        RES=$(clickhouse-client --host "${FQDN:=localhost}" --password "${PASSWORD:=}" ${PASSWORD:+--secure} --time --format=Null --query="$query" --progress 0 2>&1 ||:)
        [[ "$?" == "0" ]] && echo -n "${RES}" || echo -n "null"
        [[ "$i" != $TRIES ]] && echo -n ", "

[...]

PR #688, opened 2025-11-11
Fix Hyper scripts to perform hot runs

caetanosauer:
A recent commit aimed at "Prevent cheating by Munich-based databases" (4a2f3fa) incorrectly modified Hyper's `query.py` scripts to restart the database server on each iteration, essentially making every run a cold run. This unfairly brought our results down by a large margin.

This commit reverts the previous behavior which was already correct and there was absolutely no cheating involved.

### DIFF
--- a/hyper-parquet/query.py
+++ b/hyper-parquet/query.py
@@ -7,17 +7,17 @@ from tableauhyperapi import HyperProcess, Telemetry, Connection, CreateMode, Hyp
 
 query = sys.stdin.read()
 
-for try_num in range(3):
-    if try_num == 0:
-        # Flush OS page cache before first run of each query
-        subprocess.run(['sync'], check=True)
-        subprocess.run(['sudo', 'tee', '/proc/sys/vm/drop_caches'], input=b'3', check=True, stdout=subprocess.DEVNULL)
+with HyperProcess(telemetry=Telemetry.DO_NOT_SEND_USAGE_DATA_TO_TABLEAU) as hyper:
+    with Connection(hyper.endpoint) as connection:
+        # Hyper only supports temporary external tables, so we need to create them on every query
+        connection.execute_command(open("create.sql").read())
+        for try_num in range(3):
+            if try_num == 0:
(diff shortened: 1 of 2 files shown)


rschu1ze:
@caetanosauer There was some discussion about what "cold" runs really mean in https://github.com/ClickHouse/ClickBench/issues/667. I pushed https://github.com/ClickHouse/ClickBench/pull/692 in response. Tl;dr is that today's Hyper submission is fine - now we'll only need to migrate all other submissions one-by-one from lukewarm to cold runs for fair comparison. Allow me to close this PR - thanks.

PS: I removed "cheating" from the message of https://github.com/ClickHouse/ClickBench/pull/659. This term was unnecessarily dramatic.

caetanosauer:
@rschu1ze Thanks for your response and the update on the lukewarm definitions.

The problem I mentioned in this PR still persists, though: There are no hot runs being measured for Hyper. Last time I measured, Hyper was second place behind Umbra in the "Hot Run" category. Now we are way down below, because all 3 repetitions are run on a freshly restarted Hyper.

rschu1ze:
@caetanosauer I see, let's re-open the PR.

However, I like to clarify with you if the first run is cold or lukewarm (see the PR / issue I linked above).

In

```py
with HyperProcess(telemetry=Telemetry.DO_NOT_SEND_USAGE_DATA_TO_TABLEAU) as hyper:
    with Connection(hyper.endpoint, 'hits.hyper', CreateMode.NONE) as connection:
        for _ in range(3):
            start = timeit.default_timer()
            try:
                connection.execute_list_query(query)
```

it is not clear to me if the Hyper process survives between queries (or to be more precise: survives after three runs of a query).

We should physically terminate the database to ensure true cold runs.

caetanosauer:
The `with HyperProcess` guard starts a new process and terminates it at the end. Since the script `query.py` is invoked for each query from `run.sh` after clearing the filesystem cache, the first iteration of each query will be truly cold.

PR #679, opened 2025-11-02
update: sail 0.4.1

shehabgamin:
1. Update to Sail 0.4.1
2. The secondary runtime is now enabled by default in Sail 0.4, so I removed the redundant configuration that enabled it.

PR #691, opened 2025-11-13
Update results for sail 0.41

rschu1ze:
This is follow-up for https://github.com/ClickHouse/ClickBench/pull/679


rschu1ze:
The goal of removing two hardware configurations (out of six previously) was to make it easier to update runtimes on our end. The removed configurations are similar to other configurations, so Sail can still be compared easily with other systems.

PR #692, opened 2025-11-13
Tag runs as lukewarm

rschu1ze:
This PR addresses https://github.com/ClickHouse/ClickBench/issues/667. It introduces a distinction between lukewarm runs and cold runs. The former is considered legacy and README.md now encourages to migrate to the latter.

For the time being, all existing submissions were labeled `lukewarm-cold-runs` except
- DBaaS entries (Redshift, BigQuery, etc.),
- the databases touched by https://github.com/ClickHouse/ClickBench/pull/659 (Umbra, Hyper, CedarDB) - these have proper cold runs already.
- all entries below "versions/" and "hardware/" (they are unrelated to the official ClickBench benchmark).

We encourage PRs that migrate submissions from lukewarm to cold runs.
(We'll adjust all ClickHouse-related entries by ourselves soon).

### DIFF
--- a/datafusion-partitioned/results/c6a.2xlarge.json
+++ b/datafusion-partitioned/results/c6a.2xlarge.json
@@ -5,7 +5,7 @@
     "cluster_size": 1,
     "proprietary":   "no",
     "tuned":   "no",
-    "tags":   ["Rust","column-oriented","embedded","stateless"],
+    "tags":   ["Rust","column-oriented","embedded","stateless", "lukewarm-cold-run"],
     "load_time": 0,
     "data_size": 14737666736,
     "result": [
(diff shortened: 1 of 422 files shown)


datafusion-partitioned/results/c6a.2xlarge.json
```diff
@@ -5,7 +5,7 @@
     "cluster_size": 1,
     "proprietary":   "no",
     "tuned":   "no",
-    "tags":   ["Rust","column-oriented","embedded","stateless"],
+    "tags":   ["Rust","column-oriented","embedded","stateless", "lukewarm-cold-run"],
```
alamb, line 8: DataFusion has no state (every invocation of `datafusion-cli` is an entirely new process, and thus has no caches from the previous run) so I don't think this tagging is correct. I will make a PR to fix it

Issue #593, opened 2025-08-30
Sail: OOM on `t3a.small` machine.


alexey-milovidov:

rschu1ze:
Running sail 0.41 (added [here](https://github.com/ClickHouse/ClickBench/pull/679)) on `t3a.small` fails with

```
14221312K ........ ........ ........ ........ 98% 54.0M 2s
14254080K ........ ........ ........ ........ 98%  130M 2s
14286848K ........ ........ ........ ........ 99% 69.1M 1s
14319616K ........ ........ ........ ........ 99%  115M 1s
14352384K ........ ........ ........ ........ 99% 50.2M 1s
14385152K ........ ........ ........ ........ 99% 48.1M 0s
14417920K ........ .......                   100%  257M=2m43s

2025-11-13 14:52:31 (86.6 MB/s) - ‘hits.parquet’ saved [14779976446/14779976446]

Traceback (most recent call last):
  File "/ClickBench/sail/./query.py", line 3, in <module>
    from pysail.spark import SparkConnectServer
ModuleNotFoundError: No module named 'pysail'
```

This funny error happens because pysail cannot be build earlier:

```
  note: This error originates from a subprocess, and is likely not a problem with pip.
  full command: /ClickBench/sail/myenv/bin/python3 /ClickBench/sail/myenv/lib/python3.11/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py build_wheel /tmp/tmplwdqskbg
  cwd: /tmp/pip-install-lrtuoi63/pysail_b4807127b28440b597dd4db3ec033f35
  Building wheel for pysail (pyproject.toml): finished with status 'error'
  ERROR: Failed building wheel for pysail
Failed to build pysail
ERROR: Could not build wheels for pysail, which is required to install pyproject.toml-based projects
```

... which happens because parquet cannot be build successfully:

```

     Compiling brotli v8.0.2
     Compiling seq-macro v0.3.6
     Compiling parquet v56.2.0
     Compiling hashbrown v0.14.5
     Compiling bigdecimal v0.4.9
     Compiling xz2 v0.1.7
     Compiling bon-macros v3.8.1
     Compiling libbz2-rs-sys v0.2.2
     Compiling bzip2 v0.6.1
     Compiling bon v3.8.1
     Compiling serde_bytes v0.11.19
     Compiling anyhow v1.0.100
     Compiling apache-avro v0.20.0
     Compiling sqlparser v0.58.0
[...]

PR #695, opened 2025-11-15
update: sail 0.4.2

shehabgamin:
- Updates to Sail 0.4.2 and removes installation from source
- Hopefully closes https://github.com/ClickHouse/ClickBench/issues/593
- Adds an empty `c8g.4xlarge` results file. If only four hardware configurations are allowed, would it be okay to remove `c6a.metal` in favor of `c8g.4xlarge`?

cc @alexey-milovidov @rschu1ze

### DIFF
--- a/sail-partitioned/results/c6a.2xlarge.json
+++ b/sail-partitioned/results/c6a.2xlarge.json
@@ -5,7 +5,7 @@
     "cluster_size": 1,
     "proprietary": "no",
     "tuned": "no",
-    "tags": ["column-oriented", "lukewarm-cold-run"],
+    "tags": ["column-oriented"],
     "load_time": 0,
     "data_size": 14779976446,
     "result": [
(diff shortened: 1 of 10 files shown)


alexey-milovidov:
> If only four hardware configurations are allowed

Right now we test everything on c7a.metal-48xl, c6a.metal, c6a.4xlarge, c6a.2xlarge, c6a.xlarge, c6a.large, t3a.small, c8g.metal-48xl, c8g.4xlarge (9 machines). It's ok to submit some of them, as the result will be regenerated.

rschu1ze:
@shehabgamin Before merging, I like to ask something. There was some earlier discussion about lukewarm vs. cold runs, please see https://github.com/ClickHouse/ClickBench/pull/692 and the linked bug.

I see that the Sail submission does this:

```py
server = SparkConnectServer()
server.start()
_, port = server.listening_address
```

in sail/query.py.

1. Is Sail is a persistent server which runs in the background?

2. If yes, is the server terminated between the three runs for each query?

If the answer to 2. is "no", kindly adjust the script so it does true cold runs - thanks! (I already pushed a commit which removes the "lukewarm" tag).

alexey-milovidov:
@rschu1ze, yes - I use `./collect-results.sh`, it creates results files when run from the repository and is pointed to the database with results.

Issue #678, opened 2025-10-29
Reported load times of some systems are impossibly low


lukasvogel:
Some systems report load times below the theoretical minimum.

For example, [gp2 provides ~250MB/s throughput](https://docs.aws.amazon.com/ebs/latest/userguide/ebs-volume-types.html).
With the smallest data size anyone reports being ~10 GB, any load time below 40 seconds should be phyiscally impossible (if we assume the input file to be cached which is fair imho for the big instances like c6a.metal).

Yet, several systems report far lower load times. For example, ClickHouse on c6a.metal reports just 7 seconds.
We had a quick look at the docs and it seems that ClickHouse doesn't perform `fsync`s by default, which means the file system actually spends about two minutes syncing data after the load but before the first query is executed.

So while the benchmark reports "7 seconds", the actual import process isn't complete at that point. It seems misleading to report that ClickHouse (or any other of those systems) loads the data in 7 seconds under those conditions.

By contrast, most of the other systems include the sync to disk in the load time before acknowledging the `COMMIT` to ensure crash consistency and durability, putting them at a clear disadvantage in this metric.

puzpuzpuz:
> For example, ClickHouse on c6a.metal reports just 7 seconds.

I think, the files are in the page cache when the import is run. That's because the CSV/parquet files are first downloaded and then uncompressed to the local fs, leaving the uncompressed file pages sitting in the page cache.

alexey-milovidov:
I've also found this observation.

Downloading the dataset with `wget` also brings it to the page cache. That's why the subsequent loading does not use disk IO on machines with 16 GB of memory or more.

It is okay to measure the "hot" load time this way, as it makes the difference more visible.

PR #704, opened 2025-11-26
introduce sync calls for systems not doing fsync

lukasvogel:
Added sync around data loading for ClickHouse and Ursa as their load times are below the theoretical minimum for instances where the whole dataset fits into memory (cf Issue #678) .

I would also have liked to done it for TCHouse-C and Clickhouse Cloud but I have no idea of how to do that there.

I'm also sure there are other systems that don't do syncs but have longer load times anyway - I currently don't have time to test them all, though.

Also measured that locally on CedarDB and DuckDB - didn't have any effect, as expected. If you want to, we can also include it in future commits, though!


george-larionov (approved):
LGTM, I'll merge this PR and then re-run the affected benchmarks on our end to double check. Thanks for contributing!

clickhouse/benchmark.sh
```diff
@@ -41,8 +41,22 @@ seq 0 99 | xargs -P100 -I{} bash -c 'wget --continue --progress=dot:giga https:/
...
 
-echo -n "Load time: "
-clickhouse-client --time --query "INSERT INTO hits SELECT * FROM file('hits_*.parquet')" --max-insert-threads $(( $(nproc) / 4 ))
+# Do a sync before loading the data to make sure we don't penalize the system for having to sync more than it actually wrote
+sync
+
+# Measure load time including sync to disk
+start=$(date +%s.%N)
+
+clickhouse-client \
```
rschu1ze, line 50: @george-larionov Should we simply add MT setting [fsync_after_insert](https://clickhouse.com/docs/operations/settings/merge-tree-settings#fsync_after_insert) to the CREATE TABLE definition? This seems easier (fewer code changes) and more concise.

PR #729, opened 2025-12-17
Add pg_clickhouse

theory:
It dupes the clickhouse benchmark config to load data and the Postgres config to run the queries. All queries push down to ClickHouse without error.

Resolves ClickHouse/pg_clickhouse#82.


pg_clickhouse/benchmark.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+# apt-get update -y
```
rschu1ze, line 3: l. 7 and 8 shouldn't be commented out, no?

pg_clickhouse/.gitignore
rschu1ze, line 1: Let's remove this gitignore file (other submissions also don't have gitignore files). All output log files are useful and help to diagnose issues.

pg_clickhouse/benchmark.sh
theory: Line 1 isn't commented-out; it's a [shebang line](https://en.wikipedia.org/wiki/Shebang_%28Unix%29)

Lines 7-8 are just for my reference when testing in a fresh Ubuntu container; all the other scripts assume those items are already installed.

theory: Bah, sorry, mis-read that. Lines 3-4 are for reference; lines 7-8 definitely shouldn't be commented-out. Fixed in ea700552.

pg_clickhouse/.gitignore
theory: Sure, done in 454e65e9.

pg_clickhouse/clickhouse.sh
```diff
@@ -0,0 +1,49 @@
+#!/bin/bash
+
+# Install
+
+if [ ! -x /usr/bin/clickhouse ]
+then
+    cd /tmp || exit
+    curl https://clickhouse.com/ | sh
+    sudo ./clickhouse install --noninteractive
```
rschu1ze, line 9: I think a ClickHouse installation isn't necessary (in fact, all the `sudo` calls mess with the system). The first run of `./clickhouse` will unpack the binary into the local directory. We
- can remove l. 9,
- create `compression.yaml` in subfolder `config.d/` of the local working directory (l. 20)
- remove `sudo` from l. 23 and run the database in the background (append ` &`)
- move the files in a subfolder `user_files` of the local working directory without `sudo` (l. 45). I think `chown` (l. 46) isn't needed as well.

theory: Do you want to make the same changes to [clickhouse/benchmark.sh](https://github.com/ClickHouse/ClickBench/blob/d2cee64bad50fbec37bee169944409533d2e5527/clickhouse/benchmark.sh)? I just copied that code.

rschu1ze: Hmm, okay, then such changes would have a too large blast radius.

I'll run this script tomorrow or on Friday in our automation. This will generate the results file that can then be uploaded to https://benchmark.clickhouse.com/

theory: So, this?

```patch
--- a/pg_clickhouse/clickhouse.sh
+++ b/pg_clickhouse/clickhouse.sh
@@ -6,7 +6,6 @@ if [ ! -x /usr/bin/clickhouse ]
 then
     cd /tmp || exit
     curl https://clickhouse.com/ | sh
-    sudo ./clickhouse install --noninteractive
     rm clickhouse
     cd - || exit
 fi
@@ -20,7 +19,7 @@ compression:
     " | sudo tee /etc/clickhouse-server/config.d/compression.yaml
 fi;
 
-sudo clickhouse start
+clickhouse start &
 
 for _ in {1..300}
 do
@@ -42,8 +41,7 @@ clickhouse-client < create"$SUFFIX".sql
 
 # seq 1 | xargs -P100 -I{} bash -c 'wget --continue --progress=dot:giga https://datasets.clickhouse.com/hits_compatible/athena_partitioned/hits_{}.parquet'
 seq 0 99 | xargs -P100 -I{} bash -c 'wget --continue --progress=dot:giga https://datasets.clickhouse.com/hits_compatible/athena_partitioned/hits_{}.parquet'
-sudo mv hits_*.parquet /var/lib/clickhouse/user_files/
-sudo chown clickhouse:clickhouse /var/lib/clickhouse/user_files/hits_*.parquet
+mv hits_*.parquet /var/lib/clickhouse/user_files/
 
 echo -n "Load time: "
 clickhouse-client --time --query "INSERT INTO hits SELECT * FROM file('hits_*.parquet')" --max-insert-threads $(( $(nproc) / 4 ))
```

rschu1ze (approved):

PR #355, opened 2025-04-27
Bump StarRocks to 3.4 and remove `DISTRIBUTED BY HASH(UserID)`

azat:
I was looking at some queries where StarRocks was magnitude faster, and turns out this was due to `DISTRIBUTED BY HASH(UserID) BUCKETS 192`, which looks like a hack. So I've replaced it with `DISTRIBUTED BY RANDOM BUCKETS 1`. The comparison for this change can be found [this](https://pastila.nl/?0013bfaf/cc79ca6121da4b2bafc05816cf98d3b3.html#s2LdCLWtCaDRGx4X4wP7jA==).

So after this change StarRocks became slightly slower.

Also bump the StarRocks version to latest and fix scripts here and there.

PR #732, opened 2025-12-25
update starrocks result based on 4.0.2 and c8g.48xl

before-Sunrise:


starrocks/results/c8g.metal-48xl.json
rschu1ze, line 1: There are already result json files for six system. As per this PR,
- a new configuration is added which is basically very similar to existing configuration `c7a.metal-48xl.json`
- the existing six configurations are not updated.

Can I recommend to remove `c8g.metal-48xl.json` and instead update the existing ones? (or let us do the latter)?

starrocks/create.sql
```diff
@@ -106,5 +106,5 @@ CREATE TABLE hits (
     CLID INT NOT NULL
 )
 DUPLICATE KEY (CounterID, EventDate, UserID, EventTime, WatchID)
-DISTRIBUTED BY RANDOM BUCKETS 1
+DISTRIBUTED BY HASH(UserID) BUCKETS 192
```
rschu1ze, line 109: For my understanding: Why does the partitioning scheme change? ClickBench runs on a single node, so hash partitioning probably doesn't make a difference at all.

starrocks/results/c8g.metal-48xl.json
before-Sunrise: > There are already result json files for six system. As per this PR,
> 
> * a new configuration is added which is basically very similar to existing configuration `c7a.metal-48xl.json`
> * the existing six configurations are not updated.
> 
> Can I recommend to remove `c8g.metal-48xl.json` and instead update the existing ones? (or let us do the latter)?

so if i want to add a new configuration for a  new version of starrocks, I need to update  every exsited configurations? I want to keep the c8g.metal-48xl's result, so can you help to update the other configuration based on starrocks 4.0.2?

starrocks/create.sql
before-Sunrise: > For my understanding: Why does the partitioning scheme change? ClickBench runs on a single node, so hash partitioning probably doesn't make a difference at all.

This is the most common table creation method for users in starrocks community. First, hot and cold partitioning is based on time, and then hash bucketing is performed based on users. I think this can better reflect the starrocks' performance. I also noticed the apache doris use hash bucketing  for table creation and  clickbench accepted it, so I think starrocks can use hash bucketing too

starrocks/results/c8g.metal-48xl.json
rschu1ze: New configuration `c8g.metal-48xl` has 192 vCPUs and 384 GiB.
Existing configuration `c7a.metal-48xl.json` has 192 vCPUs and 384 GiB.
Existing configuration `c6a.metal` has 192 vCPUs and 384 GiB.

These configurations have virtually the same expressiveness, keeping all of them around increases the maintenance overhead on our end. Since you like to keep `c8g.metal-48xl`, I'm going to remove the other two.

starrocks/create.sql
rschu1ze: (for reference: https://github.com/ClickHouse/ClickBench/pull/355)

PR #734, opened 2025-12-25
Use Polars streaming

ritchie46:
Hi, I am from the Polars team, and for benchmarking purposes we always recommend to run on the streaming engine. It is much faster than the in-memory engine for almost all queries.


ritchie46:
Thank you @rschu1ze!

PR #736, opened 2025-12-27
fix Polars queries

ritchie46:
After #734, I was looking at the queries, and I noticed that some are just plainly wrong (e.g. produce different output that the SQL queries), so I did a pass and ensured they produce the right output now and utilize ideomatic Polars.


ritchie46:
I see what it is. I can push a fix the day after tomorrow.

PR #743, opened 2025-12-30
update starrocks result based on 4.0.2 and c8g.48xl, modify load script

before-Sunrise:
Follow-up to https://github.com/ClickHouse/ClickBench/pull/740


before-Sunrise:
@rschu1ze modified the load script, I have checked it in aws c8g machine, could you please help to review it?

Issue #733, opened 2025-12-25
Add GizmoSQL


alexey-milovidov:
https://github.com/gizmodata/gizmosql

rschu1ze:
Hi @prmoore77 @drabastomek You seem the [main contributors](https://github.com/gizmodata/gizmosql/graphs/contributors) to GizmoSQL. Perhaps you like to submit a GizmoSQL script and result to ClickBench?

prmoore77:
hi @rschu1ze and @alexey-milovidov - I think this is a great idea.  We should overlap with DuckDB somewhat (b/c we use DuckDB engine - with an Arrow Flight SQL wrapper) - but b/c we are a server, I think there is enough differentiation to merit an independent submission.

Thanks for thinking of us!

rschu1ze:
Thanks @prmoore77 
We'll hopefully find time this or next week to review.

PR #745, opened 2025-12-31
Add GizmoSQL

prmoore77:
Closes https://github.com/ClickHouse/ClickBench/issues/733

### DIFF
--- a/gizmosql/results/c6a.4xlarge.json
+++ b/gizmosql/results/c6a.4xlarge.json
@@ -6,7 +6,7 @@
     "proprietary": "no",
     "hardware": "cpu",
     "tuned": "no",
-    "tags": ["C++","column-oriented","arrow-flight-sql","duckdb-backend","cold-run","tls-enabled"],
+    "tags": ["C++","column-oriented","arrow-flight-sql","duckdb-backend","cold-run","tls-disabled"],
     "load_time": 131.966,
     "data_size": 26924298240,
     "result": [
--- a/gizmosql/template.json
+++ b/gizmosql/template.json
@@ -7,7 +7,8 @@
     "C++",
     "column-oriented",
     "arrow-flight-sql",
-    "duckdb",
-    "cold-run"
+    "duckdb-backend",
+    "cold-run",
+    "tls-disabled"
   ]
 }


gizmosql/template.json
```diff
@@ -0,0 +1,13 @@
...
+  "system": "GizmoSQL",
+  "proprietary": "no",
+  "hardware": "cpu",
+  "tuned": "no",
+  "tags": [
+    "C++",
+    "column-oriented",
+    "arrow-flight-sql",
+    "duckdb",
+    "lukewarm-cold-run"
```
rschu1ze: Please see [here](https://github.com/ClickHouse/ClickBench/?tab=readme-ov-file#caching) for more details what warm / lukewarm and cold runs mean in the context of ClickBench.

Two comments related to that.

run.sh contains this:

```
# Execute all queries in one session (so authentication overhead is minimized)
echo "Running benchmark with $(wc -l < queries.sql) queries, ${TRIES} tries each..."

gizmosqlline \
  -u 'jdbc:arrow-flight-sql://localhost:31337?useEncryption=true&disableCertificateVerification=true' \
  -n clickbench \
  -p clickbench \
  -f "${TEMP_SQL_FILE}"
```

The script does _not_ clear the OS page cache between query runs - which it should do to qualify as a "lukewarm run". If authentication is costly, then feel free to disable it entirely (the script seems to disable certificate validation already).

The other thing is that we are moving away from lukewarm runs. New submissions ideally do cold runs right from the start. Would it be possible to kill/start the GizmoSQL docker container between query runs?

prmoore77: hi @rschu1ze - would we need to kill/restart the container between each individual query execution?

rschu1ze: You would need to do this:
- kill/restart the container, then run query 1 three times (the first execution is a cold run, the second and third execution are hot runs),
- kill/restart the container, then run query 2 three times,
- kill/restart the container, then run query 3 three times,
- etc.

prmoore77: Thanks for the clarification, I appreciate it.  Sorry for my confusion - I'm a little new to this...

prmoore77: I may switch to just using the `gizmosql_server` executable directly - to eliminate Docker from the test stack...  It will make the installation just a tad more difficult - but not too bad...

george-larionov: > Thanks for the clarification, I appreciate it. Sorry for my confusion - I'm a little new to this...

It is no problem at all! We are always happy to expand the benchmark to new systems.

> I may switch to just using the gizmosql_server executable directly - to eliminate Docker from the test stack... It will make the installation just a tad more difficult - but not too bad...

It is totally up to you, some DBs use docker (like [CedarDB](https://github.com/ClickHouse/ClickBench/blob/main/cedardb/benchmark.sh)) while others install directly (like [DuckDB](https://github.com/ClickHouse/ClickBench/blob/main/duckdb/benchmark.sh)).

gizmosql/results/c6a.4xlarge.json
```diff
@@ -0,0 +1,57 @@
...
+    "system": "GizmoSQL",
+    "date": "2025-12-31",
+    "machine": "c6a.4xlarge",
+    "cluster_size": 1,
+    "proprietary": "no",
+    "hardware": "cpu",
+    "tuned": "no",
+    "tags": ["C++","column-oriented","arrow-flight-sql","duckdb-backend","lukewarm-cold-run","tls-enabled"],
+    "load_time": 131.966,
+    "data_size": 26924298240,
```
rschu1ze, line 11: Feel free to remove the "lukewarm" tag now that you implemented proper cold runs (thanks).

prmoore77: Thanks, @rschu1ze - I've updated that now...

george-larionov:
> Hi @george-larionov - I think that the env var: HOME must be set for the extensions to install. Could you set that to: /home/ubuntu and retry?
>
>export HOME=/home/ubuntu

@prmoore77 that worked!

PR #746, opened 2026-01-05
Update Arc benchmark scripts for conditional auth support

xe-nvdk:
- Make token handling conditional: auto-detect if auth is enabled
- Get execution_time_ms from JSON response instead of logs
- Support both auth-enabled and auth-disabled Arc configurations
- Works with Arc 26.01.1+ which has auth disabled by default

### DIFF
--- a/arc/benchmark.sh
+++ b/arc/benchmark.sh
@@ -51,15 +51,9 @@ echo "  Memory limit: ${MEM_LIMIT_GB}GB (80% of ${TOTAL_MEM_GB}GB total)"
 echo ""
 
 # ============================================================
-# 3. START ARC AND CAPTURE TOKEN FROM LOGS
+# 3. START ARC AND CAPTURE TOKEN FROM LOGS (if auth enabled)
 # ============================================================
 echo "Starting Arc service..."
-
-# Check if we already have a valid token from a previous run
-if [ -f "arc_token.txt" ]; then
-    EXISTING_TOKEN=$(cat arc_token.txt)
-    echo "Found existing token file, will verify after Arc starts..."
-fi
 
 sudo systemctl start arc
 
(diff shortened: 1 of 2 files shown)


arc/benchmark.sh
```diff
@@ -78,35 +72,50 @@ for i in {1..30}; do
...
+AUTH_ENABLED=false
 
-# First, check if existing token works
-if [ -n "$EXISTING_TOKEN" ]; then
-    if curl -sf http://localhost:8000/health -H "x-api-key: $EXISTING_TOKEN" > /dev/null 2>&1; then
-        ARC_TOKEN="$EXISTING_TOKEN"
-        echo "[OK] Using existing token from arc_token.txt"
-    else
-        echo "Existing token invalid, looking for new token in logs..."
+# Try a simple query without auth - if it works, auth is disabled
```
rschu1ze, line 79: Unless enabled/disabled authentication somehow contributes to the benchmark results (I'm 99% sure that's not the case), we should rather keep the scripts as simple and dumb as possible, i.e. use a fixed auth token or disable authentication.

The PR message says:

> Works with Arc 26.01.1+ which has auth disabled by default

Auth seems disabled by default in the latest Arc version. ClickBench only benchmarks the latest version. We don't need the complexity of toggling auth or backwards compatibility with older versions.

xe-nvdk: Is not contributing in terms of results, but the script fails. So, if we keep as is, is going to fail.

Yes, is disabled, but by mistake, we shipped and arc.toml local that we used for development that has auth disabled. Today, we are shipping 26.01.2 that fix that and a hot fix in another stuff.

xe-nvdk: So, the good thing that in this case, we don't need implement this conditional check, so, I think that we can close this PR without merging.

PR #763, opened 2026-01-27
Bump sail to 0.5.2

shehabgamin:


sail-partitioned/benchmark.sh
```diff
@@ -45,11 +45,8 @@ echo "Install Python packages"
 python3 -m venv myenv
 source myenv/bin/activate
 pip install --upgrade setuptools wheel
-pip install --no-cache-dir "pysail>=0.4.2,<0.5.0"
-pip install "pyspark-client==4.0.0" \
-  "protobuf==5.28.3" \
-  "grpcio==1.71.2" \
-  "grpcio-status==1.71.2" \
+pip install --no-cache-dir "pysail>=0.4.6,<0.6.0"
```
rschu1ze: Two questions

1. Can the script please fix the `pysail` version? Right now, it installs some unspecified version between 0.4.6 and 0.6. [As far as I see](https://pypi.org/project/pysail/), the latest version in 0.4.6. With the current approach, future benchmark runs may yield different results than today's benchmark runs.

2. Looking quickly at the [sail changelog](https://docs.lakesail.com/sail/latest/reference/changelog), I couldn't find any changes related to performance compared to the [previous update](https://github.com/ClickHouse/ClickBench/pull/698) to 0.4.2. Should we wait with a ClickBench PR until the next major release (0.5.x)?

shehabgamin: @rschu1ze

1. My intention was to allow ClickBench to always use the latest patch version. Sail follows semantic versioning (`MAJOR.MINOR.PATCH`) with frequent patch releases roughly every two weeks, while minor releases happen about every two months, with `0.5` expected next week. Rather than opening a new ClickBench PR for every patch release, I thought it would be more practical to update ClickBench only on minor releases.
2. Yeah forsure, happy to wait for the next minor release. Since `0.5` is coming next week, that was the motivation for preparing this PR 🙂

shehabgamin: @rschu1ze Updated with Sail 0.5.1!

sail-partitioned/benchmark.sh
```diff
@@ -45,11 +45,8 @@ echo "Install Python packages"
 python3 -m venv myenv
 source myenv/bin/activate
 pip install --upgrade setuptools wheel
-pip install --no-cache-dir "pysail>=0.4.2,<0.5.0"
-pip install "pyspark-client==4.0.0" \
-  "protobuf==5.28.3" \
-  "grpcio==1.71.2" \
-  "grpcio-status==1.71.2" \
+pip install --no-cache-dir "pysail>=0.5.1,<0.6.0"
```
rschu1ze: Coming back to my earlier comment, in the interest of reproducible benchmarks. we should have a _fixed_ sail version here, and not an arbitrary version betwen 0.5.1 and 0.6.

shehabgamin: @rschu1ze Done!

PR #765, opened 2026-01-28
Gizmosql fixes

george-larionov:
Some fixes for the gizmosql system, now the automation should run successfully.


prmoore77:
hi @george-larionov - should I also submit a PR for a lukewarm-cold run?  Clearing the cache between every query will definitely slow things a bit...

george-larionov:
@prmoore77 it seems the results are a bit worse after clearing the cache in between each query (as required by our cold runs). Let me know if it looks wrong to you, I can always re-run.

Also, I will add the other machine results in a separate PR.

prmoore77:
In all seriousness, thanks for your help: @george-larionov - I appreciate it.

PR #771, opened 2026-01-29
Adding results for GizmoSQL benchmark on c8g.metal-48xl 

george-larionov:
Adding c8g.metal-48xl results.


prmoore77:
DuckDB prides themselves on having the #1 "Hot run" on ClickBench for open-source systems.  Would I be allowed to submit (maybe in another folder) - a hot run ?  I'm not sure if that is allowed/possible...

PR #773, opened 2026-02-01
Update Arc benchmark results for c6a.4xlarge

xe-nvdk:
Hey team, new version released, new benchmark results, I just did c6a.4xlarge, I understand that after verification, you will run in the others sizes? if not, let me know and I can update the other sizes too.


xe-nvdk:
> These two scripts implement the automated runs:
> 
> * https://github.com/ClickHouse/ClickBench/blob/main/run-benchmark.sh
> * https://github.com/ClickHouse/ClickBench/blob/main/cloud-init.sh.in
> 
> > is right away the instance is created and accessible?
> 
> I would be very surprised if the EBS volume is not available right away.

Is available, but there's initilization that occurs in the background, and what I found is when you clone the repo, and run the benchmark right away and that initialization is not completed, the values are worst.

Remember that Ubuntu AMI, use an snapshot, and how I quoted the AWS page, is confirming that during the initialization, the performance is degraded. 

> During this time, the volume being initialized might experience increased I/O latency and decreased performance. Full volume performance is achieved only once all storage blocks have been downloaded and written to the volume.

xe-nvdk:
@rschu1ze When you have a moment, please let me know if I should provide the results for the others sizes or is managed from your end? Thank you.

PR #837, opened 2026-04-29
Fix Trino REGEXP_REPLACE backreference syntax

piotrrzysko:
The replacement string used `\1` instead of `$1` ([Trino syntax](https://trino.io/docs/current/functions/regexp.html)). In Trino, `\1` is interpreted as an escape followed by the literal character `1`, so the query was producing `1` instead of the captured domain name.


findepi (approved):

piotrrzysko:
@alexey-milovidov could you please take a look at this PR?

PR #838, opened 2026-04-29
Rework UI

rotmanjanez:
I regularly look at the ClickBench website, and slowly, the selectors become more and more convoluted, and viewing the actual results becomes more annoying. This is an attempt to fix the UI experience of the online benchmark results websites.

Main Changes:
1. Move Combined Benchmark ranking to  the top. It is a bit odd going to a benchmark site and only seeing a bunch of buttons and no results. Moving the table to the top instantly shows benchmark results to first visitors and long-time viewers. Making the combined table scrollable should also improve viewing the site on mobile devices (though still far away from mobile-friendly).
2. Update System selectors: There are eleven versions of ClickHouse and nine versions of DuckDB in the system selector filter. If I am not interested in DuckDB for whatever reason, I’d like to disable/enable all of them. Many systems encode information about data format used in parenthesis behind the database name. I split those up and moved their information to the other tags that are already in place. For commonly used logically grouped tags, I added additional selectors for easy access.
3. Add RAM selector to split up the large Machine selector and shorten names.
4. Remove XS, S, M, L, XL, 2XL, 3XL, 4XL, 2XS, and XXL as filtering options. They pollute the Machine selector, are just an alias for cluster size and therefore do not add any value for filtering.

Upsides: Much easier to filter "Compare all ClickHouse variants" or "What is the best engine for Parquet", but filters like "Show me just CedarDB without Parquet and ClickHouse with Parquet" do not work anymore. I don't think this is an issue in practice tough and is worth the improvement is filtering.

I tried to make old links with hashes still work, and from my limited testing, they still work as intended.

I hope you and others like the direction of this PR.

### DIFF
--- a/index.html
+++ b/index.html
@@ -553,12 +553,38 @@ let tags = document.getElementById('selectors_tags');
 let machines = document.getElementById('selectors_machine');
 let cluster_sizes = document.getElementById('selectors_cluster_size');
 
-let unique_databases = [... new Set(data.map(elem => elem.system))].sort((a, b) => a.localeCompare(b, undefined, {numeric: true, sensitivity: 'base'}));
+function normalizeDatabase(system) {
+    return system.replace(/\s*\([^)]*\)/g, '').replace(/\s+/g, ' ').trim();
+}
+
+function tagsFromDatabaseName(system) {
+    return [...system.matchAll(/\(([^)]*)\)/g)]
+        .flatMap(match => match[1].split(','))
+        .map(tag => tag.trim())
+        .filter(tag => tag.length > 0);
+}
+
+function uniqueValues(values) {
+    return [...new Set(values)];
+}
+
+data.forEach(elem => {
+    elem.database = normalizeDatabase(elem.system);
+    elem.tags = uniqueValues([...(elem.tags || []), ...tagsFromDatabaseName(elem.system)]);
+});
+
+let unique_databases = [... new Set(data.map(elem => elem.database))].sort((a, b) => a.localeCompare(b, undefined, {numeric: true, sensitivity: 'base'}));
+
+const database_aliases = {};
+data.forEach(elem => {
+    database_aliases[elem.database] = elem.database;
+    database_aliases[elem.system] = elem.database;
+});
 
 const database_to_tags = {};
 data.forEach(d => {
-    if (!database_to_tags[d.system]) database_to_tags[d.system] = new Set();
-    (d.tags || []).forEach(t => database_to_tags[d.system].add(t));
+    if (!database_to_tags[d.database]) database_to_tags[d.database] = new Set();
+    (d.tags || []).forEach(t => database_to_tags[d.database].add(t));
 });
 
 function highlightDatabaseTags(databaseName) {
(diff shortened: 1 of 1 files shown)


rschu1ze:
@rotmanjanez I generally like the direction, thanks. Some comments:

> 1. Move Combined Benchmark ranking to the top.

Makes sense.

> Making the combined table scrollable should also improve viewing the site on mobile devices (though still far away from mobile-friendly).

Not a blocker: The (minor) problem here is that if only few databases are selected, the graph is still padded.

grafik

Can the graph height be auto-adjusted or be made configurable via some click-on-the-border-and-drag feature?

> 2. Update System selectors.

This one is a bit more controversial. My main quirk is that the "Additional filters" button is too hard to spot, IMHO.

grafik

Most users will need to click it open because extra filtering is a necessity. So we can show the other filters always (as before), i.e. no filter button.

> 3. Change default metric back to hot.

This reverts https://github.com/ClickHouse/ClickBench/pull/616 and it is the most controversial change :-) I'd say real workloads are a mix of cold and hot runtimes anyways, so "combined" seems a good default choice. Viewers can switch to hot or cold if they like to. Could I ask you to revert this change?

Thanks for the PR (you are clearly much better with HTML and JS than me ;-) )

rotmanjanez:
@rschu1ze thanks for the review. Really appreciate it, and agree with most of it.

- updated the summary table so it now does not show whitespace when only few systems are selected
- removed the additional systems button. Now all filters are always visible. I still find the additional `database`  filter very convenient to hide multiple variations, or to e.g. compare just ClickHouse on different harder, all with just one click.
- I changed the default metric back to combined. While I don't agree with the argument, the scope of the PR is UI improvements.

I am happy to address any other issues or try out suggestions, if you have any.

P.S.: I had a university deadline. I am much more available now (hours/days, not weeks ._.). Sorry for the inconvenience.

rschu1ze:
> updated the summary table so it now does not show whitespace when only few systems are selected

This interestingly introduced another slightly annoying behavior. Start with zero selected databases, then add one, e.g. "BigQuery", then add another one, e.g. "ByConity", and another one "CedarDB". Due to auto-resizing, the selector jumps away from the mouse pointer.

Also, if all databases are selected, the chart is size-capped and users need to use the scrollbar to see all bars.

What do you think about two new buttons below the "Tuned" selector where users can set the height of the chart area. Values can be numeric (1, 2, ... rows) or `auto` (--> scrollbar). I guess we should split that off into a separate PR and make this PR only about the next item.

> removed the additional systems button. Now all filters are always visible. I still find the additional database filter very convenient to hide multiple variations, or to e.g. compare just ClickHouse on different harder, all with just one click.

There are now two categories with the confusing names "Databases" and "Systems". What about "Databases" and "Database Scenario" (or: "Database Configuration")?

> I changed the default metric back to combined. While I don't agree with the argument, the scope of the PR is UI improvements.

Thanks. Please also update the original PR message above^^.

rotmanjanez:
I removed all table related changes and focused even more on the filtering. My goal was to remove duplication from the filters, by moving the information that databases put in braces. This reduced some exponential variations (Parquet, no Parquet, partitioned, not partitioned). Further, I removed some Machine tags that are aliases for cluster sizes.

I re-added the dropdown though for some of the selectors I consider rarely used. I think the additional clarity overweights the additional click for only some small number of users. I may be wrong about the frequency, but I think many of the tags add little value and should therefore be hidden.

Let me know if I could address the some of the controversy of my changes.

Thank you again for your time^^

PR #839, opened 2026-05-01
Update both duckdb-vortex variants

AdamGS:
This PR fixes the `duckdb-vortex-partitioned` setup which was very much out of date to be much closer to the updateed `duckdb-vortex` one, and updates actual results for both duckdb-vortex variants.


AdamGS:
Seems like a more thorough version of this was merged a couple of days ago, so I'm closing this PR.

PR #845, opened 2026-05-04
Fix starrocks and doris cheating at cold runs.

alexey-milovidov:


rschu1ze:
@HappenLee Can you please clarify what information the segment cache and the storage page cache in Doris store and how they work (lifecycle)? I found some scattered bits of information in the Doris docs (e.g. https://doris.apache.org/docs/3.x/admin-manual/trouble-shooting/memory-management/memory-analysis/doris-cache-memory-analysis) but it is not really well documented. In particular, is there a way to clear these caches before each first cold query? If yes, then let's do so instead. Note that "doris-parquet/run.sh" (but not "doris/run.sh") does `curl -sS http://127.0.0.1:8040/api/clear_cache/all` already does that - will that do the trick?

This PR disables the caches globally, which also impacts hot runs (and that may be unfair).

@murphyatwork I have the similar question for Starrocks. However, in the case of the data cache, the [docs](https://docs.starrocks.io/docs/data_source/data_cache_troubleshooting/) say

```
Currently, Data Cache does not provide a direct interface to clear the cached data.
```

so disabling the data cache globally seems fair. 

Can you please also explain what the block cache is doing (mentioned [here](https://docs.starrocks.io/docs/data_source/data_cache/#enable-and-configure-data-cache))? It is _not_ disabled by this PR. Should it? Can it be cleared between queries otherwise?

HappenLee:
@rschu1ze @alexey-milovidov Hello, Here are my two questions regarding this issue:

First, we consider the page cache mechanism to be reasonable. Its logic is similar to that of DuckDB's buffer pool—previously accessed disk files are pinned and cached in the queue. In real-world production environments, users actually use it this way. So why is it considered unreasonable? If that is the case, could I equally argue that DuckDB's results are unreasonable as well?

Second, if this is indeed unreasonable, shouldn’t we clarify the rules clearly and check each database's results accordingly? For closed-source databases, how can we ensure fairness and verifiability under such rules?

rschu1ze:
Nevermind, I reverted this PR, sorry for the confusion.

alexey-milovidov:
@HappenLee, @rschu1ze, Cold result should run with no caches. Otherwise, the results are non-representative.

PR #858, opened 2026-05-06
Add spark-velox entry (Spark + Velox via Apache Gluten)

alexey-milovidov:
## Summary
- Adds a `spark-velox/` entry that runs Apache Spark with the [Velox](https://velox-lib.io/) C++ vectorised execution engine, loaded via Apache Gluten's Velox backend.
- Compared with `spark-gluten/`, this entry pins the Gluten backend to `velox` explicitly via `spark.gluten.sql.columnar.backend.lib`, so the benchmark name matches the engine actually being measured (Gluten in principle also supports a ClickHouse backend).
- Memory split between Spark heap and Velox off-heap is left as in the existing accelerator entries.

## Limitations
Apache Gluten only publishes pre-built Velox bundles for `linux_amd64`. ARM hosts have to build from source, so this PR ships without ARM results — same constraint as the existing `spark-gluten/` entry.

## Test plan
- [ ] Run on an x86_64 c6a / c8g machine.
- [ ] Verify `benchmark.sh` downloads the Gluten tarball, extracts the Velox bundle, runs all 43 queries, and writes `results/<machine>.json`.


alexey-milovidov:
Same as `spark-gluten`

PR #860, opened 2026-05-07
Refactor: standard install/start/check/stop/load/query interface per system

alexey-milovidov:
Closes #793.

## Summary

- **Split each local system's monolithic `benchmark.sh` into 7 single-purpose scripts** (`install`, `start`, `check`, `stop`, `load`, `query`, `data-size`) with a stable contract, driven by a new shared `lib/benchmark-common.sh`.
- **Wrap dataframe / in-process systems** (pandas, polars-dataframe, chdb-dataframe, daft-parquet*, duckdb-dataframe, sirius) in small FastAPI servers so they fit the same start/stop/query lifecycle.
- **88 local systems refactored**; cloud/managed systems and a handful of non-functional ones are intentionally untouched.

## Why

Previously, every system's `benchmark.sh` bundled installation, server lifecycle, dataset download, data loading, and query dispatch into one script — and `run.sh` hard-coded the per-query orchestration. There was no programmatic per-query entry point, so:

1. Tweaking the dataset, query set, or per-query behavior (e.g. restarting the system between queries to neutralize warm-process effects) required editing every system's scripts individually.
2. Building an online "run query X against system Y" service was impossible.
3. Most `run.sh` ran all 3 tries inside a single CLI invocation, so OS-cache warmth from try 1 leaked into tries 2/3.

## The new per-system interface

| Script      | Stdin     | Stdout                   | Stderr                                       | Notes |
|-------------|-----------|--------------------------|----------------------------------------------|-------|
| `install`   | -         | progress                 | progress                                     | Idempotent. Env prep + system install. |
| `start`     | -         | -                        | progress                                     | Start daemon. Idempotent. Empty/exit-0 for stateless tools. |
| `check`     | -         | -                        | progress                                     | Trivial query (e.g. `SELECT 1`). Exit 0 iff responsive. |
| `stop`      | -         | -                        | progress                                     | Stop daemon. Idempotent. |
| `load`      | -         | progress                 | progress                                     | Runs create.sql + loads data; **deletes source files** then `sync`. |
| `query`     | one query | query result, any format | **last line: fractional seconds** (`0.123`)  | Non-zero exit on failure. |
| `data-size` | -         | bytes (one integer)      | -                                            | Reports the data footprint. |

Each system's `benchmark.sh` becomes a 4-line shim that sets a couple of env vars and `exec`'s the shared driver:

```bash
#!/bin/bash
export BENCH_DOWNLOAD_SCRIPT="download-hits-parquet-partitioned"
export BENCH_RESTARTABLE=yes
exec ../lib/benchmark-common.sh
```

The shared driver runs `install → start+check → download → load (timed) → for each query: flush caches; if BENCH_RESTARTABLE=yes, stop+start; run query 3× → data-size → stop`. The output log shape (`Load time:`, `[t1,t2,t3],` per query, `Data size:`) is identical to the old `benchmark.sh`, so `cloud-init.sh.in`'s POST to play.clickhouse.com keeps working unchanged.

`BENCH_RESTARTABLE=no` for embedded CLIs (duckdb, sqlite, datafusion, …) and dataframe wrappers — restarting a single CLI/Python process between queries would dominate query time. For these, OS caches are still flushed between queries.

## Scope

**Refactored (88 systems):**

- *Server, restartable*: clickhouse, postgresql, mysql, mariadb, monetdb, druid, pinot, vertica, exasol, kinetica, heavyai, questdb, cockroachdb, elasticsearch, ydb, … and the postgres/clickhouse/mysql variants (timescaledb, citus, paradedb, postgresql-indexed, clickhouse-parquet*, clickhouse-datalake*, mysql-myisam, tidb, infobright, …)
- *Embedded CLI, not restartable*: duckdb (and variants), sqlite, datafusion (and partitioned), glaredb (and partitioned), hyper, hyper-parquet, octosql, opteryx, sail (and partitioned), drill, turso, chdb, chdb-parquet-partitioned
- *Dataframe with FastAPI wrapper, not restartable*: pandas, polars-dataframe, chdb-dataframe, daft-parquet, daft-parquet-partitioned, duckdb-dataframe, sirius
- *Spark family*: spark, spark-auron, spark-comet, spark-gluten

**Not refactored (intentionally out of scope):**

- *Cloud / managed*: alloydb, athena, aurora-{mysql,postgresql}, bigquery, clickhouse-cloud, databricks, motherduck, redshift, redshift-serverless, snowflake, hydrolix, firebolt(*), hologres, tinybird, hydra, mariadb-columnstore, pg_duckdb*, singlestore, supabase, tablespace, tembo-olap, timescale-cloud, crunchy-bridge-for-analytics, s3select, …
- *Non-functional*: csvq, dsq, locustdb (panic on first query); exasol, spark-velox (empty dirs)
- *Non-SQL or no SQL CLI*: mongodb (JS aggregation pipelines), polars (no SQL CLI; the dataframe variant is wrapped instead)

## Validated end-to-end on a 96-core / 185 GB ARM machine

| System | Data | Outcome |
|---|---|---|
| clickhouse | 14.2 GB / 100M rows | Full 43 queries × 3 tries with stop/start between queries; load 124s |
| duckdb | 20.6 GB / 100M rows | Full 43 queries × 3 tries (no restart); load 69s |
| pandas | 4.2 GB in-mem (5M-row subset) | 42/43 queries; Q43 hit a pandas lambda bug → recorded as `null` (framework's error path works) |
| sqlite | 3.9 GB (5M-row subset) | First 5 queries × 3 tries; load 68s |
| postgresql | 100M rows / 75 GB TSV | First 3 queries × 3 tries with restart; load 829s. Cold-cache spike clearly visible (135s → 7s after warmup) — confirms per-query restart actually flushes the page cache |

All 88 refactored systems pass `bash -n` and have executable bits set on the 7 scripts + benchmark.sh.

## Bug fixes surfaced during validation

- `lib/benchmark-common.sh`: `data-size` now runs **before** `stop` (clickhouse and pandas need the server up to report size).
- `clickhouse/start`: idempotent (was erroring when already running).
- `duckdb/load`, `sqlite/load`: `rm -f hits.db`/`mydb` for idempotent reruns.
- `postgresql/load`: `-v ON_ERROR_STOP=1` so COPY data errors actually fail the script instead of silently rolling back.
- `BENCH_DOWNLOAD_SCRIPT` may now be empty for systems that read directly from S3 datalakes / remote services (clickhouse-datalake*, duckdb-datalake*, chyt, …).

## Flagged for follow-up review

- `duckdb-memory` — `:memory:` semantics force a per-query reload; will inflate timings vs. the original single-process flow.
- `cloudberry`, `greenplum` — multi-phase install (reboot between phases); the shim only runs phase 1.
- `sirius` — GPU-dependent; long-lived `duckdb` CLI subprocess proxy; review the stdin/sentinel protocol.
- `paradedb*`, `pg_ducklake`, `pg_mooncake` — Docker container created in `install` then `docker cp` in `load` (small divergence from the original `docker run -v ...` due to the lifecycle order: `start` runs before `download`).

## Test plan

- [x] `bash -n` on all 88 systems' scripts
- [x] clickhouse: full 43-query benchmark.sh on 100M-row real data
- [x] duckdb: full 43-query benchmark.sh on 100M-row real data
- [x] pandas: 43-query benchmark.sh on a 5M-row subset
- [x] sqlite: abbreviated benchmark.sh on a 5M-row subset
- [x] postgresql: abbreviated benchmark.sh on full 100M-row data
- [x] Smoke-run on a fresh c6a.metal/equivalent VM via cloud-init for a representative system from each family before merging
- [x] Verify play.clickhouse.com log-ingestion sink continues to parse the output for at least one production benchmark run


clickhouse-datalake-partitioned/load
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+# Data lives in S3 (100 partitioned parquet files); create.sql defines a
+# remote-backed ENGINE = S3 table. Each query script invocation runs create.sql
+# inline before the query, so nothing is loaded locally here.
+./clickhouse local --path . --query="$(cat create.sql); DROP TABLE hits"
```
alexey-milovidov: No, let's create the table once. (here and for non-partitioned option)

alexey-milovidov: Done in 98c67c43 — `load` now creates the table once via clickhouse-local's persistent `--path` metadata, and `query` just runs the query. Tested end-to-end: `./install && ./load` then `echo 'SELECT count() FROM hits' | ./query` returns 99,997,497 against the persisted table.

clickhouse/query
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+# Reads a SQL query from stdin, runs it via clickhouse-client.
+# Stdout: query result (default format).
+# Stderr: query runtime in fractional seconds on the last line (from --time).
+# Exit non-zero on error.
+set -e
+
+query=$(cat)
+clickhouse-client --time --query="$query"
```
alexey-milovidov: It also accepts queries in stdin, so it will be even simpler.

alexey-milovidov: Change all ClickHouse entries accordingly.

clickhouse/start
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+set -e
+
+# Idempotent: if already up, do nothing.
```
alexey-milovidov: Remove this as `start` is already idempotent.

clickhouse/query
alexey-milovidov: Done — clickhouse-client reads stdin directly, dropped the explicit cat. Verified success and error paths still produce timing on the last stderr line / non-zero exit respectively.

clickhouse/start
alexey-milovidov: Done in this commit. I dropped the explicit idempotency check in `clickhouse/start` and made the shared driver (`lib/benchmark-common.sh`) tolerant of non-zero exits from `./start` — `bench_check_loop` is the authoritative readiness signal. This pattern works for any system whose start command leaves the post-condition satisfied even when it returns non-zero.

alexey-milovidov:
## 1. Core refactor — per-system script interface (`1f352adcc` and follow-ups)

Replaced each system's monolithic `benchmark.sh` with a stable set of single-purpose hooks driven by `lib/benchmark-common.sh`:

```
install     env prep + system install (idempotent)
start       start daemon (idempotent; no-op for stateless tools)
check       trivial query / health probe — exit 0 iff responsive
stop        stop daemon (idempotent; no-op for stateless tools)
load        runs create.sql + ingests data + sync
query       SQL on stdin → result on stdout, fractional seconds on
            stderr's last line; non-zero exit on error
data-size   data footprint in bytes
```

88 systems migrated. Each system's `benchmark.sh` is now a 4-line shim setting env vars and `exec`'ing the lib. Output format (`[t1,t2,t3]`, `Load time:`, `Data size:`) preserved exactly so the `sink.parser` materialized view keeps working.

## 2. Library (`lib/benchmark-common.sh`)

- **Restart cycle ordering**: `./stop → wait_stopped → drop_caches → ./start → check_loop` (in that order — flushing before stop leaves mmap-backed engines pinning pages).
- **`BENCH_DURABLE` semantics** (renamed from `BENCH_RESTARTABLE`):
  - `yes` (default): data is on disk → existing cold cycle.
  - `no` (in-process servers): data in process RAM, so re-run `./load` after each restart and roll the load wall-clock into the first ("cold") try.
- **`bench_load` flushes** with `sync` inside the timed window, so `load_time` is the honest "data is on disk" wall-clock for every system.
- **Concurrent-QPS test** (`bench_concurrent_qps`): 10 workers × 600 s (configurable) fire queries against the running engine. Each worker uses a SHA-256-derived deterministic permutation of the query indices, so connection 3 hits the same query order on every engine. A side watchdog revives the engine mid-test without halting workers. Emits `Concurrent QPS:` and `Concurrent error ratio:`.
- **Partial-load detection**: after `./load`, require `./data-size ≥ 5 GB` to catch silent OOM-mid-COPY.
- **Spark progress-bar parse fix**: `tr '\r' '\n'` before numeric grep — Spark's `\r`-redrawn progress bar was concatenating the timing line onto the bar.
- **`./check` failure on timeout** now surfaces the actual stderr instead of a generic message.

## 3. Per-system fixes (by failure mode)

### arm64 readiness
- Arch-aware downloads for `databend`, `octosql`.
- `opteryx` 0.26.1 → 0.26.8 (gained aarch64 wheel).
- Fail-fast on arm64 for `hyper{,-parquet}`, `doris{,-parquet}`, `citus`, `pgpro_tam` (no upstream arm64 artifact).

### Q29 (`REGEXP_REPLACE`) rewrites
For engines that don't ship the function:
- `mariadb-columnstore` — rewrite using `SUBSTRING_INDEX` + `REPLACE`.
- `sqlite` — rewrite using `INSTR` + `SUBSTR` + `IIF`.

### Server health probes
- Presto (4 variants): replaced `grep "SERVER STARTED" docker logs` with HTTP `/v1/info` probe.
- CrateDB `./check` waits for shard recovery (`sys.shards` state), not just the wire protocol.
- Firebolt now `grep`s the body for "Firebolt is ready" so "Cluster not yet healthy" 200-with-JSON-error doesn't pass the wait loop.

### Server lifecycle
- GizmoSQL `stop_gizmosql` polls until the process actually dies (was `wait` on a non-child).
- `start_gizmosql` retries with bounded waits on DuckDB lock conflict.

### Elasticsearch 9.4 bootstrap
- `discovery.type: single-node` (the new bootstrap requires `cluster.initial_master_nodes` to match an existing `node.name`; old config silently fell through to red status, every bulk write 503'd).

### OOM diagnostics
- 8 server-load scripts (`pandas`, `polars-dataframe`, `chdb-dataframe`, `duckdb-{dataframe,memory}`, `daft-parquet{,-partitioned}`, `sirius`) now capture the curl body and emit the server error before `set -e`-exiting, instead of producing 266 s "successful" empty runs.

### `pg_ducklake`
- DuckDB `memory_limit` bumped to 80 % of host RAM so CTAS doesn't OOM at the 4 GB container default.

### Dataframe systems
- `chdb-dataframe`, `daft-parquet{,-partitioned}`, `duckdb-dataframe`: dropped the SQL whitelist + `_make_runner` factory — `/query` now executes the body directly.
- `pandas`, `polars-dataframe`: server `/query` `eval()`s a Python expression from `queries.py` (one expression per line, parallel to `queries.sql`); no more hardcoded `(sql, lambda)` table.

### Spark
- Pinned `pyspark==3.5.5` (4.0 broke the shared `query.py`).

### Misc
- Drill — use `--run=` instead of stdin.
- Pinot, Sirius, CockroachDB — bumped `BENCH_CHECK_TIMEOUT` for cold-start.
- Kinetica — fixed timing parse.
- ByConity — docker compose v2 install.
- datafusion-vortex — submodule init.
- Druid — JDK 17.
- ParadeDB — switch to `pg_search` after `pg_lakehouse` removal.

## 4. New tags / taxonomy

- **`"in-memory"`** added to all dataframe / Python-HTTP-server entries (templates + historical results).
- **`"DuckDB derivative"`** added to 17 systems × all their results (`duckdb*`, `pg_duckdb*`, `pg_ducklake`, `motherduck`, `gizmosql`, `sirius`, `pg_mooncake`, `crunchy-bridge-for-analytics`) — mirrors the existing `"ClickHouse derivative"` convention.
- **`"serverless"`** removed from local-storage systems (was incorrectly applied to embedded engines like chDB).
- **`"lukewarm-cold-run"`** dropped from post-refactor results and the corresponding templates — the new `BENCH_DURABLE` machinery makes cold genuinely cold.
- **`"stateless"`** reapplied uniformly to Parquet / data-lake / Spark / Sail / Polars entries.

## 5. Cloud-init / operator side (`run-benchmark.sh`, `cloud-init.sh.in`)

- `export HOME=/root` in cloud-init so child install/query scripts find a home directory (fixes XDG-convention tools, GizmoSQL installer, DuckDB extension cache).
- Configurable global timeout via `timeout=` env (default bumped 20000 s → 36000 s).
- 16 GB swapfile added on launch (helps the row-oriented loads on c6a.4xlarge).
- `run-instances` retries every 60 s indefinitely on transient capacity/quota errors:
  - `InsufficientInstanceCapacity`
  - `VcpuLimitExceeded`
  - `InstanceLimitExceeded`
  - `MaxSpotInstanceCountExceeded`

  Real config errors still propagate immediately.

## 6. Database schema (`prepare-database.sql`)

- Added `concurrent_qps Nullable(Float64)` and `concurrent_error_ratio Nullable(Float64)` to `sink.results`.
- `sink.parser` MV extracts both via `toFloat64OrNull` and includes them in the per-row `output` JSON.
- `generate-results.sh` now skips entries of the form `{"error": "<url>"}`.

## 7. Dashboard (`index.html`, `generate-results.sh`)

- Per-row remove control.
- Measurement-date column in summary rows.
- "Stateless" exclusion filter in the "load" metric.
- Tag-hover highlights.
- More compact selectors at the top.

## 8. Error / timeout result records

- New minimal `{"error": "<github-issue-url>"}` form for failed runs (vs. reconstructing partial timings).
- 4 CedarDB timeout records added under `cedardb/results/20260510/`.

## 9. Result data

- ~2,200 result JSONs touched by the tag sweeps (lukewarm-cold-run, serverless, in-memory, DuckDB derivative, stateless).
- ~500 new result files from post-refactor benchmark runs.

## 10. Helper artifact

- `rerun-within-10x-clickhouse.sh`: 452 invocations across 59 systems whose c6a.metal total runtime is within 10× of clickhouse — output of the rerun-prioritisation query.

PR #863, opened 2026-05-07
GizmoSQL: switch from gizmosqlline (Java) to gizmosql_client

prmoore77:
## Summary

Switches the GizmoSQL ClickBench scripts from the Java-based `gizmosqlline` SQLLine client to the new C++ `gizmosql_client` shell that ships in the same release archive as `gizmosql_server`.

- Drops the `openjdk-17-jre-headless` apt install and the separate `gizmosqlline` binary download — `gizmosql_client` is already in `gizmosql_cli_linux_${ARCH}.zip`.
- Replaces `gizmosqlline -u <jdbc-uri> -n -p -f` invocations with `gizmosql_client --file ...`, which reads connection parameters from the `GIZMOSQL_HOST` / `GIZMOSQL_PORT` / `GIZMOSQL_USER` / `GIZMOSQL_PASSWORD` env vars natively.
- In `run.sh`, prepends `.timer on` and `.mode trash` to each per-query batch so the client emits one `Run Time: X.XXs` line per statement (to stderr) and skips rendering result rows. stderr is merged into stdout for log capture.
- Updates the result parser regex in `benchmark.sh` from sqlline's `rows? selected ([0-9.]+ seconds)` to `Run Time: [0-9.]+s`.

See https://docs.gizmosql.com/#/client for full client usage.

## Test plan

- [x] Smoke-tested locally with `gizmosql_server` / `gizmosql_client` v1.22.4 on a small in-memory table; the `Run Time: X.XXs` parser produced the expected `[t1, t2, t3],` output rows.
- [ ] Full ClickBench run on a Linux host (the cache-drop and apt steps are Linux-only).

### DIFF
--- a/gizmosql/benchmark.sh
+++ b/gizmosql/benchmark.sh
@@ -4,8 +4,8 @@
 export HOME=/home/ubuntu
 
 # Install requirements
-apt-get update -y
-apt install unzip netcat-openbsd -y
+sudo apt-get update -y
+sudo apt-get install -y unzip netcat-openbsd
 
 # Detect architecture (maps x86_64->amd64, aarch64->arm64)
 ARCH=$(uname -m)
@@ -15,10 +15,12 @@ elif [ "$ARCH" = "aarch64" ]; then
     ARCH="arm64"
 fi
 
-# Install the GizmoSQL server and client (gizmosql_client is the CLI shell)
+# Install the GizmoSQL server and client (gizmosql_client is the CLI shell) into a local directory
+mkdir -p ./bin
 curl -L -o gizmosql.zip "https://github.com/gizmodata/gizmosql/releases/latest/download/gizmosql_cli_linux_${ARCH}.zip"
-unzip gizmosql.zip
-mv gizmosql_server gizmosql_client /usr/local/bin/
+unzip -o gizmosql.zip -d ./bin
+chmod +x ./bin/gizmosql_server ./bin/gizmosql_client
+export PATH="$PWD/bin:$PATH"
 
 # Source our env vars and utility functions for starting/stopping gizmosql server
 . util.sh
--- a/gizmosql/benchmark.sh
+++ b/gizmosql/benchmark.sh
@@ -5,7 +5,7 @@ export HOME=/home/ubuntu
 
 # Install requirements
 apt-get update -y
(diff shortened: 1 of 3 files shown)


gizmosql/benchmark.sh
```diff
@@ -5,7 +5,7 @@ export HOME=/home/ubuntu
 
 # Install requirements
 apt-get update -y
-apt install openjdk-17-jre-headless unzip netcat-openbsd -y
+apt install unzip netcat-openbsd -y
```
rschu1ze: L. 7 and 8 need `sudo`, please check the other invocations of `apt` in the repository.

gizmosql/benchmark.sh
```diff
@@ -15,40 +15,25 @@ elif [ "$ARCH" = "aarch64" ]; then
     ARCH="arm64"
 fi
 
-# Server setup Install
+# Install the GizmoSQL server and client (gizmosql_client is the CLI shell)
```
rschu1ze: L. 21: Please don't copy binaries into `/usr/local/bin/`. Instead, please copy to a local directory and modify `$PATH` so that it finds the binary.

gizmosql/benchmark.sh
prmoore77: hi @rschu1ze - this should be fixed now.

gizmosql/benchmark.sh
prmoore77: hi @rschu1ze - this should be fixed now.

gizmosql/run.sh
```diff
@@ -24,6 +24,10 @@ for query in "${queries[@]}"; do
     # Start the GizmoSQL server
     start_gizmosql
 
+    # Enable timer and discard result rows (we only care about Run Time)
```
rschu1ze, line 27: Please change l. 22 to 

```
echo 3 | sudo tee /proc/sys/vm/drop_caches > /dev/null
```

I got `tee: /proc/sys/vm/drop_caches: Permission denied` errors locally.

rschu1ze (approved):

gizmosql/run.sh
prmoore77: Done - thanks!

prmoore77:
🚀🚀🚀🚀

PR #864, opened 2026-05-07
Add InfluxDB 3 Core entry

alexey-milovidov:
## Summary
- Adds an `influxdb/` benchmark entry targeting [InfluxDB 3 Core](https://docs.influxdata.com/influxdb3/core/) — the open-source, SQL-capable (DataFusion) build of InfluxDB.
- `load.py` streams `hits.tsv` to `/api/v3/write_lp`. All 105 columns are stored as fields (no tags), with a unique row-index nanosecond timestamp so points don't merge. Field names are lowercased so standard CamelCase ClickBench queries resolve under DataFusion's identifier-case folding.
- `queries.sql` is the standard ClickBench set; only Q19 and Q43 are adapted (`CAST(EventTime AS TIMESTAMP)`) since `EventTime` is stored as a string field.
- Removes InfluxDB from the README `TODO` list.

## Validation
- Verified the install + start + create-db + load + query flow on a 1000-row sample of `hits.tsv`. All 43 queries returned three timings each — no nulls, no errors. Spot-checked Q19 (`extract(minute ...)`), Q29 (`REGEXP_REPLACE`), and Q43 (`DATE_TRUNC('minute', ...)`) — all returned sensible rows.
- Full 100M-row load not run on this branch — that is best done on a benchmark VM. At ~1 KB per line-protocol point this will take some hours; results to be added after a real run.

## Test plan
- [ ] Run `benchmark.sh` on a `c6a.4xlarge` VM and capture `Load time` / `Data size` / 43 query timings.
- [ ] Add `influxdb/results/c6a.4xlarge.json`.


alexey-milovidov:
Data loading is painfully slow.

alexey-milovidov:
The problem: InfluxDB does not work at all.

```
2026-05-08T19:31:51.501856Z  INFO influxdb3_query_executor: executing sql query database=hits query=SELECT COUNT(*) FROM hits; params=None
2026-05-08T19:32:39.121739Z ERROR influxdb3_server::http: Error while handling request error=query error: error while planning query: order_union_sorted_inputs
caused by
Internal error: should not perform a file scan of overlapping ranges within same file.
This issue was likely caused by a bug in DataFusion's code. Please help us to resolve this by filing a bug report in our issue tracker: https://github.com/apache/datafusion/issues method=POST path=/api/v3/query_sql content_length=Some("75") client_ip=127.0.0.1
```

> Based on the research, this is a known assertion in InfluxDB 3's own physical optimizer (core/iox_query/src/physical_optimizer/sort/regroup_files.rs::group_same_file_sources) that fires when planning detects    
  overlapping byte/row ranges within one Parquet file. There is no public fix; v3.9.2 still ships the assertion. Our load.py deterministically triggers it: 16 parallel writers all submit batches whose timestamps  
  interleave through [TS_BASE, TS_BASE+N), so every WAL segment ends up containing the same broad time range, and on flush the resulting Parquet files have near-identical [min_time, max_time]. The sort-pushdown   
  pass then trips when it tries to regroup their scan ranges.                                                                                                                                                        
                                                                                                                                                                                                                    > A reload is unavoidable — the Parquet layout on disk is what's broken, and queries will keep failing until the data is rewritten in a layout the planner can handle.                                               
                                                                                
> The fixes worth considering:                                                                                                                                                                                       
                                                                                
> 1. Disjoint time-range chunks (recommended) — split the 100M rows into N contiguous chunks. For each chunk, load all batches in parallel, then kill -TERM the server (drain WAL → Parquet) before starting the next
   chunk. Each Parquet file now covers only one chunk's time range, so files don't overlap. Keeps throughput, predictable layout.
> 2. Single-writer load — set WORKERS = 1. Simple, but ~30× slower according to the existing comment in load.py.                                                                                                     
> 3. Try a different InfluxDB version — research found no version known to be immune, so this is a gamble.

xe-nvdk:
Datafusion is already here, that is the engine that InfluxDB uses and can be a good start comparison.

Another approach can be, downloading the Parquet, and move it to data dir, but I'm not sure if InfluxDB require any kind of registration of parquet/table before is good to query.

alexey-milovidov:
It does not work on c6a.4xlarge - the first query gives OOM.

> Datafusion is already here, that is the engine that InfluxDB uses and can be a good start comparison.

Yes, as InfluxDB is based on Datafusion, we can assume that it is no better than Datafusion. But I still want to have it on ClickBench to see how much worse it is.

PR #865, opened 2026-05-07
Add Quickwit entry

alexey-milovidov:
## Summary

Adds a benchmark entry for [Quickwit](https://quickwit.io), a Rust-based search engine for log analytics built on Tantivy. Modeled on the existing Elasticsearch entry.

Quickwit exposes an Elasticsearch-compatible REST API but no SQL endpoint, so each ClickBench query is hand-translated to ES DSL in \`queries.json\`. Loading goes through \`/api/v1/_elastic/hits/_bulk\`; querying through \`/_search\`.

19 of the 43 queries are not expressible in Quickwit's ES API and are recorded as \`null\`:

| Reason | Queries |
|---|---|
| \`COUNT(DISTINCT)\` — no \`cardinality\` aggregation | 5, 6, 9, 10, 11, 12, 14, 23 |
| Substring \`LIKE '%…%'\` — leading wildcards rejected, no \`wildcard\`/\`regexp\` query | 21, 22, 24 |
| \`ORDER BY\` on text field — sort by text not supported | 26, 27 |
| Scripted/runtime fields, \`REGEXP_REPLACE\`, \`CASE WHEN\`, integer arithmetic in aggregations | 19, 28, 29, 30, 36, 40 |

The remaining 24 queries were validated against a 1M-row sample on a single c6a-class node (aarch64). The full 100M-row run was not executed in this PR; \`results/\` is empty pending a full benchmark run.

## Test plan
- [x] \`benchmark.sh\` installs Quickwit 0.8.2 and starts the server (verified, aarch64 + x86_64 via \`uname -m\`)
- [x] Index creation from \`index_config.yaml\` succeeds (105 fields)
- [x] \`load.py\` ingests via the ES bulk API (~17K docs/s for the 1M sample)
- [x] All 43 entries in \`queries.json\` parse as valid JSON; \`run.sh\` emits 24 timings + 19 \`null\`s
- [ ] Full 100M-row load and \`results/c6a.4xlarge.json\` (follow-up)


alexey-milovidov:
From @melvynator:

> Quickwit supports cardinality aggregation. https://quickwit.io/docs/main-branch/reference/aggregation#cardinality

alexey-milovidov:
> Cardinality is in v0.9 (unreleased) — only the binary distribution is for v0.8.2.

alexey-milovidov:
But we can use the unreleased version.

melvynator:
> But we can use the unreleased version.

Well, v0.8.2 is 2 years old. And the project has not made major release since the Datadog acquisition, there is no point in comparing a project that is 2 years old, so the master build is the only viable and reasonable options. 

https://github.com/quickwit-oss/tantivy/pull/2337

alexey-milovidov:
Switched to the nightly build.

PR #883, opened 2026-05-08
Add Hyrise

### LINKED ISSUE #751: Add Hyrise
https://github.com/hyrise/hyrise

alexey-milovidov:
## Summary
- Add a ClickBench entry for [Hyrise](https://github.com/hyrise/hyrise), a research in-memory column-oriented DBMS from HPI.
- Hyrise has no upstream binaries and needs gcc-15 / clang-20, so the build runs inside a multi-stage Dockerfile (ubuntu:25.04 builder → slim runtime image with `hyriseServer`, `hyriseConsole`, `libhyrise_impl.so`, `libjemalloc.so.2`). LTO is disabled by default: GCC's `-flto=auto` link spawns one `lto1` per core and OOMs a 32 GB machine.
- Load: hits.csv is split into 5M-row pieces which are imported sequentially (`COPY ... FROM` appends to an existing table), because Hyrise's CSV import reads the whole file into one `std::string` and materializes the table unencoded before dictionary-encoding it (~200 GB peak for the monolithic 75 GB file). Column types come from `hits.csv.json` next to each data file.
- The first load exports the table to Hyrise's binary format via `hyriseConsole` (`export` bypasses the MVCC `Validate` operator, so the snapshot keeps the encoded segments; SQL `COPY TO` would materialize it unencoded). All cold-cycle reloads (`BENCH_DURABLE=no`) import that snapshot, which streams chunk-by-chunk and regenerates statistics.
- Data size is reported from `meta_segments.estimated_size_in_bytes` since Hyrise has no on-disk persistence.
- Hyrise's SQL is limited: no `LENGTH`, `REGEXP_REPLACE`, `DATE_TRUNC`, or `OFFSET`. `./query` keys off psql's exit code so those queries (Q28, Q29, Q39–Q43) are recorded as `null` instead of stealing the timing line that psql still prints.

Closes #751

## Test plan
- [x] Loaded a 1000-row CSV slice; `SELECT COUNT(*)`, `MIN/MAX(EventDate)`, `AVG(UserID)`, `COUNT(DISTINCT UserID)`, `EXTRACT(MINUTE FROM EventTime)` all return expected results
- [x] All 43 queries run end-to-end: 36 produce timings, 7 produce `null` (Q28/29 — unsupported functions, Q39–Q43 — `OFFSET`/`DATE_TRUNC`)
- [x] Output format matches ClickBench expectations: `Load time:` line, `Data size:` line, 43 lines of `[t1,t2,t3],`
- [x] Piecewise-load + binary-snapshot design verified against Hyrise master sources (`csv_parser.cpp`, `import.cpp`, `binary_writer.cpp`/`binary_parser.cpp`, `console.cpp`, `sql_translator.cpp`)
- [ ] Full 100M-row run — needs a memory-rich machine (see the analysis in the comments); the estimated in-memory table size very likely exceeds c6a.4xlarge's 32 GB


alexey-milovidov:
Analyzed the failed run and pushed three fixes.

**Why the run failed:** the Docker build died linking `libhyrise_impl.so` — GCC's `-flto=auto` spawns one `lto1` per core, and 16 of them exhausted the 32 GB + 16 GB swap of c6a.4xlarge (`g++: fatal error: Terminated signal terminated program lto1` is earlyoom's SIGTERM). Fixed by defaulting the existing `NO_LTO` build arg to `TRUE` (ee02575).

**Two more failures were waiting behind the build fix**, both found by reading Hyrise's sources and fixed in this push:

1. The runtime image was missing `libboost-container1.83.0`/`libboost-date-time1.83.0`, which `libhyrise_impl.so` links dynamically — `hyriseServer` would not have started (1bea24d).
2. The monolithic `COPY hits FROM '/data/hits.csv'` cannot work at 100M rows: Hyrise's CSV import reads the entire file into one `std::string` and materializes the whole table as unencoded `ValueSegments` before encoding — roughly **200 GB peak** for the 75 GB CSV. `./load` now splits the CSV into 5M-row pieces and imports them sequentially (`COPY ... FROM` appends to an existing table), capping the transient overhead at ~13 GB above the growing encoded table. Since `BENCH_DURABLE=no` re-runs `./load` on all 43 cold cycles, the first load also exports an encoded binary snapshot via `hyriseConsole` (its `export` command bypasses the MVCC `Validate` wrapper that would materialize the snapshot unencoded), and every reload streams that snapshot back instead of re-parsing 75 GB of CSV (8896065).

**Machine-size caveat:** even with the piecewise load, the peak is bounded below by the final dictionary-encoded in-memory table, which I estimate at 30–50 GB — most likely above what c6a.4xlarge (32 GB RAM + 16 GB swap) can hold. I'd suggest running this on a memory-rich machine via the manual workflow (e.g. c6a.metal); the c6a.4xlarge PR run will probably fail at load with an earlyoom kill, and if so the standard machine should get an `{"error": ...}` result.

PR #886, opened 2026-05-09
Quickwit optimizations

alexey-milovidov:


alexey-milovidov:
None of the optimizations gives meaningful improvements.

PR #893, opened 2026-05-10
GizmoSQL: add support for specifying storage version in util.sh

prmoore77:
In this PR, I am updating the GizmoSQL server start line to use the latest version for the database file format.

### DIFF
--- a/gizmosql/util.sh
+++ b/gizmosql/util.sh
@@ -12,6 +12,7 @@ start_gizmosql() {
     nohup gizmosql_server \
         --username ${GIZMOSQL_USER} \
         --database-filename clickbench.db \
+        --storage-version latest \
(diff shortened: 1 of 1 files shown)


rschu1ze:
Wondering why the server doesn't use the latest format version by default?

prmoore77:
> Wondering why the server doesn't use the latest format version by default?

Hi @rschu1ze - DuckDB does the same thing also - they default to an older file format.  We were using their defaults.  I just shipped GizmoSQL server v1.26.0 - which added the ability to set the --storage-version arg - like the DuckDB CLI has.

See this line in their ClickBench script:
https://github.com/ClickHouse/ClickBench/blob/8cefe9d34c70656680ed4a95010b83381894f334/duckdb/benchmark.sh#L12

PR #836, opened 2026-04-28
Add ParticleDB ClickBench results

nohup23:
Adds ParticleDB ClickBench results on c8g.metal-48xl.\n\nSummary:\n- System: ParticleDB\n- Machine: c8g.metal-48xl\n- Result date: 2026-03-29\n- Includes load time from hits.parquet\n- Tags: Rust, column-oriented, HTAP, Arrow\n- Not tagged in-memory\n\nThe generated data row places ParticleDB at #1 on the current Combined list in this checkout.

PR #895, opened 2026-05-11
Add ZigHouse

donge:
## Summary
- Add ZigHouse as a ClickBench entry using a fixed Linux x86_64 release binary.
- Include runner scripts compatible with ClickBench's shared benchmark driver.
- Include AWS `c6i.4xlarge` results from the 2026-05-11 run.

## Notes
The result uses `c6i.4xlarge` because `c6a.4xlarge` was unavailable in the AWS China regions used for this run.

Release binary:
https://github.com/donge/zighouse/releases/tag/v0.1.0-clickbench

SHA256:
5a779eacf87082eeeb13b336d6f798a0399593c29b5a586ab67202104396dc83


rschu1ze:
Is there a homepage or GitHub repository for ZigHouse? https://github.com/ClickHouse/ClickBench/pull/836 made me a bit sceptical about submissions that only come with a download but not homepage/repository.

alexey-milovidov:
I'm in favor of accepting everything that works at least occasionally. I like looking at experimental, obscure, discontinued, historical, and even crackpot databases: https://presentations.clickhouse.com/2018-unknown_developers_reissue/

Just take care when testing - use only isolated machines.
We can do forensic analysis as well :)

donge:
Thanks for your feedback. I will try to make the submission satisfy the project requirements as much as possible.
ZigHouse is still experimental at this stage, so let me briefly explain what it is.

I have been a long-time ClickHouse user, and I learned a lot from ClickHouse's design while using it. ZigHouse is an attempt to build a lightweight single-node analytical database inspired by column-oriented execution, vectorized data processing, and SIMD-friendly kernels.

I also used DuckDB as an important reference point when comparing the design and behavior. At the current stage, ZigHouse is closer to a non-general-purpose, embedded DuckDB-like analytical engine rather than a mature database system.

The main target use case is lightweight embedded analytics on structured wide tables, especially time-related analytical workloads with scans, aggregations, Top-K queries, and unique/distinct-style computations.
The repository is here:
https://github.com/donge/zighouse

The slides looks fun, let me check and update the progress, Thank you all

donge:
Here is a simple homepage https://donge.github.io/zighouse/

alexey-milovidov:
The scripts work.

alexey-milovidov:
Added a result here: https://github.com/ClickHouse/ClickBench/pull/898

donge:
Thanks Alexey, truly impressive with your productivity as CTO coder

alexey-milovidov:
ZigHouse hardcodes the set of ClickBench queries and can't process any other query, so we will have to remove it.

donge:
Sure,  ZigHouse is still being built bottom-up right now. 

My goal is to set the fastest possible baseline so we don't see any performance drops when integrating generic path later on. this time only have an executor for now, without a complete SQL AST parser. 
I will make it work with common SQL next time. anyway, thanks for providing a project like this—it sets a great common benchmark.

PR #896, opened 2026-05-11
Refactor clickhouse-cloud to use the common script interface

alexey-milovidov:
## Summary

- Splits `clickhouse-cloud/benchmark.sh` into per-step scripts (`install`, `start`, `stop`, `check`, `load`, `query`, `data-size`) and exec's `lib/benchmark-common.sh`, matching `chyt` / `clickhouse-web`. Run-loop moves out of the deleted `run.sh`.
- Picks up the concurrent QPS / error-ratio measurement automatically. `collect-results.sh` now parses the new `Load time:` / `Data size:` / `Concurrent QPS:` log shape and emits `concurrent_qps` / `concurrent_error_ratio` in the result JSON.
- Adds `BENCH_MANAGED=yes` opt-in to `lib/benchmark-common.sh`. When set, the cold cycle collapses to a single `./check` — local OS page-cache flushes and `stop`/`start` are no-ops for SaaS engines anyway, and skipping the 60 s `wait_stopped` poll saves ~43 min of dead time per run (and the same again before the QPS window). `clickhouse-cloud` and `chyt` both opt in; `start` / `stop` for `clickhouse-cloud` are no-ops since the service is provisioned externally by `cloud-api.sh`.

## Test plan

- [ ] `bash -n` clean on all new scripts and on `lib/benchmark-common.sh`.
- [ ] Run `cloud-api.sh` end-to-end against a real cloud service and confirm: load completes, all 43 `[t1,t2,t3],` rows appear, `Concurrent QPS:` / `Concurrent error ratio:` lines appear, and `collect-results.sh` produces a JSON that includes `concurrent_qps` and `concurrent_error_ratio`.
- [ ] Verify wall-time savings vs previous behavior (cold cycles should be ~instant per query, not ~60 s).
- [ ] Spot-check a chyt run to confirm `BENCH_MANAGED=yes` doesn't regress its existing flow.


alexey-milovidov:
Does not work well from GitHub actions.

PR #901, opened 2026-05-14
Update ZigHouse

donge:
## Update ZigHouse

Update the zighouse entry per review feedback in #901. Switch to a generic `v1.0.3` release binary with direct Parquet import.

### Changes from review feedback

- **Generic binary** — downloads `v1.0.3` from [donge/zighouse/releases](https://github.com/donge/zighouse/releases) instead of a ClickBench-specific binary
- **Parquet import** — `import-parquet --format=generic` writes directly to a local column store, no HTTP server dependency
- **BENCH_RESTARTABLE=no** — ZigHouse is an embedded CLI, so cold-cycle restarts are skipped (just `drop_caches`)
- **generic-smoke.sh** — runs after the main benchmark (not inside `./load`) to demonstrate generic SQL without inflating `load_time`
- **BENCH_CONCURRENT_DURATION=0** — single-process engine skips QPS test

### Architecture

ZigHouse parses SQL with a native Zig recursive-descent parser, builds an IR plan, and executes via a pipeline-parallel executor. All 43 ClickBench queries run through the same generic SQL pipeline — there is no separate ClickBench-specific fast path.

### Results on c6i.4xlarge (100M rows)

| Metric | Value |
|--------|-------|
| load_time | 1366.64 s |
| data_size | 75.78 GB |
| best-of-3 warm query sum (43 q) | 46.82 s |
| null queries | 1 (Q20 segfault) |

The generic store format is deliberately simple (no compression, no marks, no primary index), resulting in larger storage but correct end-to-end import + query. Compression is on the roadmap.

### Install

Binary: https://github.com/donge/zighouse/releases/tag/v1.0.3
SHA256: `4f63353bfc9b09a649fb54c77c447a113b4bf2b331957aa65d4b33648ef56d5b`


alexey-milovidov:
Could you please sign a CLA?

donge:
Sure, done.

zighouse/install
```diff
@@ -1,8 +1,8 @@
 #!/bin/bash
 set -e
 
-: "${ZIGHOUSE_VERSION:=v0.1.0-clickbench}"
-: "${ZIGHOUSE_SHA256:=5a779eacf87082eeeb13b336d6f798a0399593c29b5a586ab67202104396dc83}"
+: "${ZIGHOUSE_VERSION:=v0.2.0-clickbench}"
```
rschu1ze: Can we you please change this to download a generic ZigHouse binary like `v1.0.2` from the [release page](https://github.com/donge/zighouse/releases) instead of a ClickBench-specific binary?

zighouse/README.md
```diff
@@ -4,7 +4,7 @@ ZigHouse is an experimental analytical database binary written in Zig.
 
 This ClickBench entry uses the published Linux x86_64 benchmark binary from:
 
-https://github.com/donge/zighouse/releases/tag/v0.1.0-clickbench
+https://github.com/donge/zighouse/releases/tag/v0.2.0-clickbench
```
rschu1ze: Please see my other comment about the binary.

zighouse/generic-smoke.sh
```diff
@@ -0,0 +1,36 @@
+#!/bin/bash
```
rschu1ze, line 1: The script isn't called from anywhere, therefore it doesn't run automatically. Could you please call it from `zighouse/load`?

rschu1ze (changes requested):

zighouse/install
donge: done

PR #902, opened 2026-05-14
Update QuestDB to 9.3.5

puzpuzpuz:
This fixes DNFs in Q27 and Q28. The reason for the DNFs is that the `length_bytes()` SQL function was shipped in 9.3.2 while clickbench installation still used 9.3.1, so the queries were not aligned with QuestDB version in the benchmark. Also, `query.timeout.sec` configuration option was replaced with `query.timeout`, so the script wasn't increasing the default query timeout properly.


puzpuzpuz:
I'll include the new result files tomorrow.

**Update**. The results are now included.

puzpuzpuz:
@rschu1ze could you take a look at this one?

puzpuzpuz:
@alexey-milovidov could you take a look at this one?

puzpuzpuz:
Is the benchmark still maintained?

puzpuzpuz:
@george-larionov could you take a look at this PR? It's pending for a while already.

puzpuzpuz:
@rschu1ze thanks for the review and double-checking the submission. It looks like your results match what we had sans the `concurrent_qps` metrics which seems to be removed.

PR #906, opened 2026-05-15
Add Frigatebird

alexey-milovidov:
## Summary

- Adds a `frigatebird/` ClickBench recipe for [Frigatebird](https://github.com/Frigatebird-db/frigatebird), an embedded columnar SQL database written in Rust (push-based Volcano execution, morsel parallelism, LZ4 + O_DIRECT storage).
- `./load` streams `hits.parquet` through a small pyarrow script (`parquet_to_inserts.py`) into the Frigatebird REPL as batched `INSERT INTO hits VALUES (...)` statements — Frigatebird has no `COPY` / Parquet / CSV ingest path.
- `create.sql` collapses all integer widths to `BIGINT` and `DATE` to `TIMESTAMP` (Frigatebird's type system has no narrower forms), and uses the mandatory `ORDER BY (CounterID, EventDate, UserID, EventTime, WatchID)`.
- `./query` measures runtime with bash built-in `time` since the CLI has no built-in timer.

## Notes

- `parquet_to_inserts.py` emits negative integers as quoted strings to work around Frigatebird's `INSERT` planner rejecting `UnaryOp { Minus, Number }` literals; the column-type coercion path parses them back to `i64`.
- Frigatebird's SQL surface doesn't include `EXTRACT`, `REGEXP_REPLACE`, `LENGTH`/`STRLEN`, `CASE`, etc., so several queries will fail at parse/plan time and land as `null` in the results JSON.
- In smoke testing, Frigatebird's TEXT decompressor panics with `failed to decompress page payload: string is not valid utf8` on the non-UTF-8 bytes that the hits dataset's text columns contain. The recipe is wired up so the upstream behaviour on the full dataset is reproducible; expect many or all queries to be `null` until upstream stabilises ingest/scan for non-UTF-8 strings.

Resolves #809

## Test plan

- [ ] `./install && ./benchmark.sh` on a fresh Ubuntu 24.04 VM
- [ ] Confirm any queries that succeed have plausible timings; remaining queries surface as `null`


alexey-milovidov:
It was updated last time 6 months ago. But I love collecting obscure and abandoned databases. Maybe we have a chance to make it work.

alexey-milovidov:
@nubskr, FYI

PR #935, opened 2026-05-20
Replace shell expansion with find to avoid ARG_MAX

tsg:
While workin on #933 , I hit this small issue, basically hitting ARG_MAX on the shell expansion. This replaces it with `find` to avoid shell expansion.

Likely only an issue on macOS, it doesn't happen when running against the EC2 server. So just a small QoL improvement.

### DIFF
--- a/generate-results.sh
+++ b/generate-results.sh
@@ -12,7 +12,8 @@ FIRST=1
 
 # Build "<system>/<basename> <full-path>" lines, then keep the last (latest)
 # row per key — sorted ascending by date, since YYYYMMDD sorts lexically.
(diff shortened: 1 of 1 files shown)


alamb:
I hit the same thing as well (and made basically the same change)
- https://github.com/ClickHouse/ClickBench/pull/949

alamb (approved):
This (or https://github.com/ClickHouse/ClickBench/pull/949)  would be a helpful change to allow others to test / submit results for ClickBench

rschu1ze:
Closed in favor of https://github.com/ClickHouse/ClickBench/pull/949

Thanks for sending a PR anyways @tsg

PR #943, opened 2026-06-01
fix(datafusion): Update docs, skip concurrent_qps, Skip re-downloading partitioned files

alamb:
# Rationale

While trying to gather results for a new DataFusion release, I could not complete the benchmark scripts because the new concurrent QPS test caused the `bash benchmark.sh` run to be OOM-killed by the kernel.

I also hit two rough edges:

1. `datafusion-partitioned/benchmark.sh` now re-downloads the partitioned Parquet source files on each run, which makes it harder to iterate on the scripts locally.
2. The DataFusion documentation still referred to the old script layout.

# Changes

This PR addresses those issues and lets me run the DataFusion benchmarks again:

1. Disable the concurrent QPS test by default for DataFusion entries, since `datafusion-cli` is configured here for single-user/single-process benchmark runs.
2. Hard-link the partitioned Parquet files into the `partitioned/` directory instead of moving them, so reruns do not re-download the source files.
3. Update the DataFusion README to point to the new `install` script when changing the DataFusion version.

I plan to submit a separate PR with updated results once the DataFusion release is finalized.

# Question: should contributors still submit updated results?

It is not clear to me after the recent refactors whether you prefer contributors to run and submit benchmark results as part of release-update PRs, or whether you prefer to use the ClickBench automation to regenerate the numbers.

The DataFusion results were recently updated in:

- https://github.com/ClickHouse/ClickBench/pull/840
- https://github.com/ClickHouse/ClickBench/pull/844

The benchmark scripts were also substantially refactored in:

- https://github.com/ClickHouse/ClickBench/pull/860

Updating ClickBench results takes me several hours for each DataFusion release, and I would like to decrease that. I looked into using `./run-benchmark.sh`, which appears to be the automation entry point, but from an external contributor perspective:

1. It posts logs to the ClickHouse Play `sink` user, and I do not know how to retrieve the generated result JSON from that pipeline.
2. The supporting pieces appear to be in `prepare-database.sql` and `collect-results.sh`, but the setup for using them with a contributor-owned ClickHouse instance is not documented.

If contributors are expected to run the benchmarks themselves, it would be helpful to document the recommended end-to-end workflow for launching the machines and collecting the generated JSON files.

### DIFF
--- a/datafusion-partitioned/benchmark.sh
+++ b/datafusion-partitioned/benchmark.sh
@@ -3,4 +3,7 @@
 export BENCH_DOWNLOAD_SCRIPT="download-hits-parquet-partitioned"
 export BENCH_DURABLE=yes
 export BENCH_RESTARTABLE=no
+# skip concurrent_qps tests by default as datafusion-cli is configured
+# for single user/single process usage
+export BENCH_CONCURRENT_DURATION="${BENCH_CONCURRENT_DURATION:-0}"
 exec ../lib/benchmark-common.sh
--- a/datafusion/README.md
+++ b/datafusion/README.md
@@ -25,7 +25,7 @@ All with no EBS optimization and no instance store.
 2. Wait for the status checks to pass, then ssh to EC2: `ssh ubuntu@{ip}`
 3. `git clone https://github.com/ClickHouse/ClickBench`
 4. `cd ClickBench/datafusion`
-5. `vi benchmark.sh` and modify the following line to target the DataFusion version
+5. `vi install` and modify the following line to target the DataFusion version
 
     ```bash
     git checkout 53.1.0
--- a/datafusion/benchmark.sh
+++ b/datafusion/benchmark.sh
@@ -3,4 +3,7 @@
 export BENCH_DOWNLOAD_SCRIPT="download-hits-parquet-single"
 export BENCH_DURABLE=yes
 export BENCH_RESTARTABLE=no
+# skip concurrent_qps tests by default as datafusion-cli is configured
+# for single user/single process usage
+export BENCH_CONCURRENT_DURATION="${BENCH_CONCURRENT_DURATION:-0}"
 exec ../lib/benchmark-common.sh
--- a/datafusion-partitioned/load
+++ b/datafusion-partitioned/load
@@ -1,9 +1,14 @@
 #!/bin/bash
 # datafusion queries the parquet files via an external table at LOCATION
 # 'partitioned' (see create.sql). The shared bench_download fetches the
-# parquet files into CWD; move them into the expected subdir.
+# parquet files into CWD; link them into the expected subdir.
+#
+# Note: don't move them to avoid re-downloading each time
 set -e
 
 mkdir -p partitioned
-mv hits_*.parquet partitioned/ 2>/dev/null || true
+for f in hits_*.parquet; do
+    ln -s -f "../$f" "partitioned/$f"
+done
+
 sync
--- a/datafusion-partitioned/load
+++ b/datafusion-partitioned/load
@@ -1,14 +1,14 @@
 #!/bin/bash
 # datafusion queries the parquet files via an external table at LOCATION
 # 'partitioned' (see create.sql). The shared bench_download fetches the
-# parquet files into CWD; link them into the expected subdir.
+# parquet files into CWD; hard-link them into the expected subdir.
 #
 # Note: don't move them to avoid re-downloading each time
 set -e
 
 mkdir -p partitioned
 for f in hits_*.parquet; do
-    ln -s -f "../$f" "partitioned/$f"
+    ln -f "$f" "partitioned/$f"
 done
 
 sync


datafusion/benchmark.sh
```diff
@@ -3,4 +3,7 @@
 export BENCH_DOWNLOAD_SCRIPT="download-hits-parquet-single"
 export BENCH_DURABLE=yes
 export BENCH_RESTARTABLE=no
+# skip concurrent_qps tests by default as datafusion-cli is configured
+# for single user/single process usage
+export BENCH_CONCURRENT_DURATION="${BENCH_CONCURRENT_DURATION:-0}"
```
alamb, line 8: When I ran the rewritten script on a c6a.4xlarge, the kernel repeatedly killed `datafusion-cli`

```shell
$ sudo dmesg 
...
[ 2474.473278] oom-kill:constraint=CONSTRAINT_NONE,nodemask=(null),cpuset=/,mems_allowed=0,global_oom,task_memcg=/user.slice/user-1000.slice/user@1000.service/tmux-spawn-15c64bd8-c60d-4843-bf48-c65eac5ced5e.scope,task=datafusion-cli,pid=19689,uid=1000
[ 2474.473439] Out of memory: Killed process 19689 (datafusion-cli) total-vm:18148748kB, anon-rss:7049888kB, file-rss:2872kB, shmem-rss:0kB, UID:1000 pgtables:15880kB oom_score_adj:0
```

alamb: The newly rewritten "benchmark-common.sh" appears to include new code to do a "concurrency" test which launches a bunch of concurrent `datafusion-cli` processes

https://github.com/ClickHouse/ClickBench/blob/a377499b950b17b2d96b79625375d57b18a36634/lib/benchmark-common.sh#L385-L418

I don't think this type of test makes sense for a stateless engine like DataFusion and doesn't reflect how people use this. Each `datafusion-cli` process thinks it has the resources of the entire machine, and doesn't coordinate between each other (by design). 

Systems built with DataFusion absolutely can and do implement various multi-user resource management policies, but each picks the policy that is best suited for its requirements. DataFusion does not mandate any particular one.

If we want to include a qps number for `datafusion-cli` the runner should measure how fast the query can be run one at a time (set concurrency to 1)

alamb: Fascinatingly there are recent results checked in for a c6a.4xlarge that have a `qps_concurrency` measurement:
https://github.com/ClickHouse/ClickBench/blob/a377499b950b17b2d96b79625375d57b18a36634/datafusion-partitioned/results/20260511/c6a.4xlarge.json#L12-L13

These appear to be added in https://github.com/ClickHouse/ClickBench/commit/6d5ee0a68573bbb481fdb8f799cc3348a9bb0634 by @alexey-milovidov, as part of PR https://github.com/ClickHouse/ClickBench/pull/860 and I don't see any comments / additional context of how it was generated) 🤔 

Given that the kernel just OOM kills my runner, I don't know how to reproduce this number. Thus I think it is better and more accurate to disable the qps measurement for DataFusion and report "null" instead

datafusion/README.md
```diff
@@ -25,7 +25,7 @@ All with no EBS optimization and no instance store.
 2. Wait for the status checks to pass, then ssh to EC2: `ssh ubuntu@{ip}`
 3. `git clone https://github.com/ClickHouse/ClickBench`
 4. `cd ClickBench/datafusion`
-5. `vi benchmark.sh` and modify the following line to target the DataFusion version
+5. `vi install` and modify the following line to target the DataFusion version
```
alamb, line 28: #860 changes the script that need to be updated

datafusion-partitioned/benchmark.sh
```diff
@@ -3,4 +3,7 @@
 export BENCH_DOWNLOAD_SCRIPT="download-hits-parquet-partitioned"
 export BENCH_DURABLE=yes
 export BENCH_RESTARTABLE=no
+# skip concurrent_qps tests by default as datafusion-cli is configured
```
alamb, line 6: See rationale below

george-larionov:
Hi @alamb, let me try to answer some of your questions.
> Question: should contributors still submit updated results?

Yes, since we cannot track changes in every benchmarked system and when it needs to be updated. The current flow (contributors submit updated results, maintainers reproduce and merge them) should still be followed.

> If contributors are expected to run the benchmarks themselves, it would be helpful to document the recommended end-to-end workflow for launching the machines and collecting the generated JSON files.

This is fair. After the recent refactor, not all the docs are consistent. I will take this as a TODO item. The `./run-benchmark.sh` script is **our** automation entry point, which is why you cannot access the sink it dumps the results into. for now please continue to use the old workflow, AKA `./benchmark.sh`, for getting updated results.

datafusion/benchmark.sh
george-larionov: It looks like https://github.com/ClickHouse/ClickBench/blob/main/cloud-init.sh.in#L20 adds earlyoom and a 16gb swapfile, which is why those runs succeeded for Alexey and not for you on a bare instance. Either way I agree this is not a useful measurement for this kind of DB. We will discuss internally how to make it more useful.

george-larionov: Created https://github.com/ClickHouse/ClickBench/issues/946 to track this.

george-larionov (approved):
The changes LGTM.

alamb:
Thank you for the review and merge @george-larionov 🙏

alamb:
> Yes, since we cannot track changes in every benchmarked system and when it needs to be updated. The current flow (contributors submit updated results, maintainers reproduce and merge them) should still be followed.

Here is an update PR for DataFustion
- https://github.com/ClickHouse/ClickBench/pull/948

Issue #946, opened 2026-06-12
concurrent_qps measures swap-thrashing, not throughput, for single-process engines (introduced in #860)


george-larionov:
## Summary

The `concurrent_qps` test introduced in #860 produces misleading, non-reproducible numbers for single-process / in-process engines (embedded CLIs and dataframe-server systems). For these systems it spawns N independent full-machine processes that collectively oversubscribe RAM, so on memory-constrained instances the metric measures **swap-thrashing throughput**, not query throughput — and whether it completes at all depends on the runner environment rather than the engine.

This surfaced in #943 (DataFusion), but the root cause is in `lib/benchmark-common.sh` (`bench_concurrent_qps`) and affects a whole class of systems.

## Evidence

`bench_concurrent_qps` launches `BENCH_CONCURRENT_CONNECTIONS` (default **10**) workers for `BENCH_CONCURRENT_DURATION` (default **600s**). For `BENCH_DURABLE=no` engines like `datafusion-cli`, each worker's `./query` spawns a fresh process that assumes it owns the whole machine. 10 × ~7 GB RSS ≈ 70 GB against 32 GB RAM.

Checked-in DataFusion (single) results, `concurrent_qps` vs. machine RAM:

| Machine | vCPU / RAM | concurrent_qps | err |
|---|---|---|---|
| c6a.xlarge | 4 / 8 GB | 0.025 | 0 |
| c6a.2xlarge | 8 / 16 GB | 0.042 | 0 |
| c6a.4xlarge | 16 / 32 GB | 0.11 | 0 |
| c8g.4xlarge | 16 / 32 GB | 0.092 | 0 |
| c6a.metal | 192 / 384 GB | 4.017 | 0 |
| c8g.metal-48xl | 192 / 384 GB | 5.388 | 0 |
| c7a.metal-48xl | 192 / 384 GB | 5.888 | 0 |

(`datafusion-partitioned` shows the same shape: 0.025–0.097 on the small boxes, 4.9–7.0 on the metals.)

The QPS tracks **RAM**, not core count — the metric on the sub-RAM machines is reporting how fast the box can page, not how fast the engine runs queries. `err=0` everywhere, so nothing crashed in the automated runs; the processes just crawled through swap.

## Why the same instance type gives different outcomes

[...]

PR #947, opened 2026-06-15
Skip concurrent-QPS test for single-process engines, null stale values

george-larionov:
## Summary

The concurrent-QPS test added in #860 (`bench_concurrent_qps` in `lib/benchmark-common.sh`) fires `BENCH_CONCURRENT_CONNECTIONS` workers (default 10) for `BENCH_CONCURRENT_DURATION` seconds (default 600). For **single-process / fork-per-query engines** — where each `./query` spawns a fresh full-machine process with no shared scheduler — this doesn't measure throughput: the 10 processes just oversubscribe RAM. On memory-constrained machines it produces swap-thrash numbers (QPS that tracks RAM, not cores) and can OOM-kill the whole run, taking the main results down with it (the symptom in #943).

This PR disables the test for that class of engines and cleans up the stale numbers it already produced. See #946 for the full analysis.

## Changes

1. **Skip the test in 37 single-process shims** — set `export BENCH_CONCURRENT_DURATION="${BENCH_CONCURRENT_DURATION:-0}"` (override-preserving, matching the form in #943, so it can be re-enabled with `BENCH_CONCURRENT_DURATION=600 ./benchmark.sh`). Covers the embedded CLIs, their parquet/datalake/vortex variants, and the Spark family.
2. **Null the stale result values** — 257 result files across those 37 systems have `concurrent_qps`/`concurrent_error_ratio` set to `null` (minimal diff: only those two value lines per file change; JSON revalidated).
3. **Short-circuit `bench_concurrent_qps` when duration ≤ 0** — emit `null` for both fields immediately instead of running the cache-flush / restart / worker setup that all no-op out anyway.
4. **Document the rule** in the `BENCH_CONCURRENT_DURATION` lib docs and the README "How To Add a New Result" section.

## Selection rule

Skip when the system has **no shared server** — `./start` launches no daemon and `./query` invokes the engine fresh per query. Kept for:

- **Daemons** (clickhouse, postgres, mysql, trino, druid, …) — one server schedules the concurrency, which is exactly what the test should measure.
- **In-process server wrappers** (`pandas`, `polars`, `*-dataframe`, `daft`, `sirius`) — one shared process, no 10× duplication.
- **Remote backends** (`chyt`, …) — queries hit a shared remote service.

A few systems (`kinetica`, `monetdb`, `questdb`, `yugabytedb`) have an idempotent `exit 0` `start` that short-circuits when the server is already up; these were confirmed to be real servers and are **not** skipped.

## Caveat — this is an educated guess

The 37-system selection is derived from reading each system's `start`/`query` scripts, **not** from empirically running the concurrent test and observing OOM vs. sustained throughput. It's a confident classification, but it's still a judgment call, and the cleanest version of this would have maintainers **opt their own system out of the metric case-by-case** — they know whether their engine shares a server better than a script-reading heuristic does. If you'd prefer that model, this PR can instead serve as the documented rule + the bulk cleanup of the existing bad data, with the per-shim flag treated as a default that owners can correct.

## Related

- #946 — issue with the full analysis (swap-thrashing evidence, the RAM-vs-cores table)
- #943 — the DataFusion-specific workaround that surfaced this; this PR generalizes it
- #860 — introduced the concurrent-QPS test


george-larionov:
@alexey-milovidov I had claude go through and opt out of the concurrent-qps metric for relevant databases. However without running the tests to see if they produce OOMs, I can't really make sure these are 100% correct, do you think I should do that, or merge it?

PR #948, opened 2026-06-18
Update Results for DataFusion version to 54.0.0

alamb:
- Part of https://github.com/apache/datafusion/issues/21505

This updates ClickBench with results for DataFusion 54 - https://datafusion.apache.org/blog/output/2026/06/12/datafusion-54.0.0/

Contains updates for:

## `datafusion`

- [x] **c6a.xlarge** — x86_64 (AMD) — 4 vCPU
- [x] **c6a.2xlarge** — x86_64 (AMD) — 8 vCPU
- [x] **c6a.4xlarge** — x86_64 (AMD) — 16 vCPU — the canonical/reference size
- [x] **c6a.metal** — x86_64 (AMD) — bare metal
- [x] **c7a.metal-48xl** — x86_64 (AMD) — bare metal, 192 vCPU
- [x] **c8g.4xlarge** — arm64 (Graviton4) — 16 vCPU
- [x] **c8g.metal-48xl** — arm64 (Graviton4) — bare metal, 192 vCPU

## `datafusion-partitioned`

- [x] **c6a.xlarge** — x86_64 (AMD) — 4 vCPU
- [x] **c6a.2xlarge** — x86_64 (AMD) — 8 vCPU
- [x] **c6a.4xlarge** — x86_64 (AMD) — 16 vCPU — the canonical/reference size
- [x] **c6a.metal** — x86_64 (AMD) — bare metal
- [x] **c7a.metal-48xl** — x86_64 (AMD) — bare metal, 192 vCPU
- [x] **c8g.4xlarge** — arm64 (Graviton4) — 16 vCPU
- [x] **c8g.metal-48xl** — arm64 (Graviton4) — bare metal, 192 vCPU

I generated these numbers using [aws_clickbench.txt](https://github.com/user-attachments/files/29132774/aws_clickbench.txt), with the help of Claude Code:

PR #949, opened 2026-06-19
Fix generate-results.sh: avoid "Argument list too long"

alamb:
## Problem
While working on https://github.com/ClickHouse/ClickBench/pull/948 I hit a problem regenerating `data.generated.js`. Specifically, running  `bash generate-results.sh` fails locally with:

```
generate-results.sh: line 15: /bin/ls: Argument list too long
```

This is because `ls */results/*/*.json`, expands **every** result path onto a single command line which exceeds the shell's `ARG_MAX`. 

There are a huge number of files in this repo now (`clickhouse-cloud` alone has tens of thousands) so I don't know how the data.generated.js that is checked in was generated. 

## Fix

Use `find` to walk the tree instead of expanding a glob:

```sh
LANG="" find */results -mindepth 2 -maxdepth 2 -name '*.json'
```

### DIFF
--- a/generate-results.sh
+++ b/generate-results.sh
@@ -12,7 +12,9 @@ FIRST=1
 
 # Build "<system>/<basename> <full-path>" lines, then keep the last (latest)
 # row per key — sorted ascending by date, since YYYYMMDD sorts lexically.
-LANG="" ls -1 */results/*/*.json \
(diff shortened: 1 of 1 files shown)


rschu1ze (approved):

rschu1ze:
Verified to work correctly under Linux and macOS.

Thanks @alamb

Issue #891, opened 2026-05-10
CedarDB: infinite loop with 100% CPU consumption after restart.


alexey-milovidov:
Only the trivial `SELECT count(*) FROM hits` finished, and after restarting and when running the second query, it hangs:
```
CREATE TABLE
Load time: 528.138
[0.191,0.012,0.010],
```
while consuming 192 * 100% of CPU cores on `c6a.metal`.

lukasvogel:
Should be solved by #897

rschu1ze:
Haven't seen this bug in https://github.com/ClickHouse/ClickBench/pull/897

Issue #543, opened 2025-07-12
Umbra: OOM on loading data


alexey-milovidov:
It can't load data on a machine with 16 GB of memory or less.

rschu1ze:
@toschmidt @lukasvogel

ChrisWint:
Thanks for reporting, we'll look into it

ChrisWint:
Created an issue for that in our Bugtracker to keep track of it: https://github.com/cedardb/issues/issues/20

ChrisWint:
@alexey-milovidov @rschu1ze This should be fixed for CedarDB in the latest release

rschu1ze:
For CedarDB, it's fixed indeed (I tested on c6a.2xlarge).

For Umbra, the issue persists. I tried on a `latest` machine (the submission previously used 25.7, [fix](https://github.com/ClickHouse/ClickBench/pull/696))

```
15466496K ........ ........ ........ ........ 97% 98.3M 4s
15499264K ........ ........ ........ ........ 97% 92.8M 4s
15532032K ........ ........ ........ ........ 97% 87.0M 4s
15564800K ........ ........ ........ ........ 97% 88.6M 3s
15597568K ........ ........ ........ ........ 98%  122M 3s
15630336K ........ ........ ........ ........ 98% 95.2M 3s
15663104K ........ ........ ........ ........ 98% 95.8M 2s
15695872K ........ ........ ........ ........ 98% 95.2M 2s
15728640K ........ ........ ........ ........ 99% 95.6M 2s
15761408K ........ ........ ........ ........ 99% 91.6M 1s
15794176K ........ ........ ........ ........ 99% 92.0M 1s
15826944K ........ ........ ........ ........ 99% 91.8M 1s
15859712K ........ ........ ........ ........ 99% 80.6M 0s
15892480K ........ ........ .......          100%  122M=2m43s

2025-11-15 17:36:58 (95.4 MB/s) - ‘hits.tsv.gz’ saved [16298506510/16298506510]

Unable to find image 'umbradb/umbra:latest' locally
latest: Pulling from umbradb/umbra
c62190a9ab61: Pull complete
61f0b5ee5e6b: Pull complete
d92623c9d7fc: Pull complete
83809684defe: Pull complete
02dbafc72a99: Pull complete
986972f530da: Pull complete
72680096bf27: Pull complete
4f4fb700ef54: Pull complete
[...]

PR #951, opened 2026-06-19
update umbra and add umbra s3

toschmidt:
Updates Umbra and improves runtimes by loading the data from parquet and using compression.

This also resolves the memory usage problem on smaller instances reported in #543 and the infinite loop during the concurrent measurements in #891.

Add a new Umbra variant that stores the data on S3 instead of the EBS device. Similar to ClickHouse (web), which also uses S3 as a storage backend.


umbra/create.sql
```diff
@@ -103,7 +103,123 @@ create table hits (
     hasgclid              smallint     not null,
     refererhash           bigint       not null,
     urlhash               bigint       not null,
-    clid                  integer      not null,
-    primary key (counterid, eventdate, userid, eventtime, watchid)
-);
-copy hits from '/data/hits.tsv' with (format text);
+    clid                  integer      not null
+) with (compression=zstd);
```
rschu1ze, line 107: Looks like the primary key definition is gone. Was this intentional?

rschu1ze, line 107: Re `compression = zstd`: I will not oppose this but it slightly violates the spirit of the benchmark. The [rules](https://github.com/ClickHouse/ClickBench/#installation-and-fine-tuning) say:

```
It's better to use the default settings and avoid fine-tuning. Configuration changes can be applied if it is considered strictly necessary and documented.

Fine-tuning and optimization for the benchmark are not recommended but allowed. In this case, add results for the vanilla configuration and tunes results separately (e.g. 'MyDatabase' and 'MyDatabase-tuned')
```

Running without `compression=zstd` or making zstd Umbra's default will be the preferred option.

umbra/create.sql
```diff
@@ -103,7 +103,123 @@ create table hits (
...
+
+-- Ingest from the Athena parquet rather than the TSV. Umbra's COPY-from-
```
rschu1ze: It's fine to keep the comment l. 110-117 around but nobody will miss it if you remove it :-)

umbra/start
```diff
@@ -8,68 +8,12 @@ fi
...
-    vm.max_map_count=1048576 || true
-
 sudo docker run -d --name umbradb \
     -v "$(pwd)/db:/var/db" \
     -v "$(pwd)/data:/data" \
     -p 5432:5432 \
+    --privileged \
     --ulimit nofile=1048576:1048576 \
     --ulimit memlock=-1:-1 \
+    -e ASYNCIO=0 \
```
rschu1ze, line 17: L. 18 deserves a comment.

umbra/start
```diff
@@ -8,68 +8,12 @@ fi
...
-#     ceiling that the kernel will OOM on regardless of how much
-#     swap is available. Let the host kernel manage memory.
-sudo sysctl -wq vm.overcommit_memory=1 vm.swappiness=100 \
-    vm.max_map_count=1048576 || true
-
 sudo docker run -d --name umbradb \
     -v "$(pwd)/db:/var/db" \
     -v "$(pwd)/data:/data" \
     -p 5432:5432 \
+    --privileged \
```
rschu1ze, line 14: L. 8 isn't needed (the `stop` script already does this).
L. 9 can move into the `stop` script as well.

rschu1ze (changes requested):
@toschmidt I get a permission denied error when I try to push to this PR.

```
/data/ClickBench (schmidt/umbra26.06 >) $ git push umbra schmidt/umbra26.06
remote: Permission to umbra-db/ClickBench.git denied to rschu1ze.
fatal: unable to access 'https://github.com/umbra-db/ClickBench.git/': The requested URL returned error: 403
```

Could I ask you to merge from `main` and resolve the conflicts?

umbra/create.sql
toschmidt: Yes it's intentional. Enforcing the uniqueness constraint is not mandatory, so the index is not needed.

umbra/create.sql
toschmidt: Agreed, I removed the comment

umbra/create.sql
toschmidt: We provide users manual control over compression settings so they can optimize for their specific storage device, cost, and access latency. We could change this to an automated policy but this would make it less obvious for users how the data is stored.

umbra/start
toschmidt: I added a comment

toschmidt: I added the comment

umbra/start
toschmidt: 👍🏼 i moved the lines to stop

alexey-milovidov:
Could you please allow pushes to your branch? It will be needed for the automation.
Or just merge the latest main, which includes the required automation.

toschmidt:
> Could you please allow pushes to your branch? It will be needed for the automation. Or just merge the latest main, which includes the required automation.

I rebased the branch, but I can't enable edits from maintainers. The fork is owned by an organization, and apparently GitHub does not support this option (https://github.com/orgs/community/discussions/5634). I can add you guys to our fork to allow edits, or do you know another way?

alexey-milovidov:
Umbra works, but Umbra S3 doesn't.

rschu1ze:
The problem is that the S3 variant requires a static AWS access key to reproduce results. We can use our own but it will leak somewhere in publicly available log files (like [here](https://github.com/ClickHouse/ClickBench/pull/951#issuecomment-5049978651)).

Would it be possible to omit -s3 for now?

toschmidt:
Yes, we can omit umbra-s3. It's possible to run it manually, the Readme contains details how to create and destroy the S3 bucket. For an automated run, I would not enable it as the bucket could be leaked and it requires permissions to create buckets in ec2 instance.

Do you want to remove it entirely or just exclude from the automated benchmarks?

toschmidt:
I removed umbra-s3 and will do a follow up pull request.

PR #952, opened 2026-06-20
Fix and document the Tinybird benchmark runner

fallintoplace:
## Summary

- replace the invalid placeholder authorization header with required `TINYBIRD_TOKEN` and `TINYBIRD_HOST` environment variables
- fail on shell, HTTP, malformed JSON, and non-numeric timing errors
- document the Tinybird CLI setup, data import, pipe generation, token creation, and benchmark invocation
- use the host selected by `tb auth -i`, preserve runner failures when logging with `tee`, and ignore local credentials and virtual environments

## Why

`AUTH_HEADER=<TOKEN>` is parsed as shell redirection, so the runner exits before sending any benchmark requests. The previous instructions also did not cover the setup needed to create the 43 pipe endpoints and run the benchmark.

## Validation

- `bash -n tinybird/run.sh`
- verified missing token and host values fail before any request
- mocked successful Tinybird responses and verified the output contains 43 arrays with three numeric timings each
- verified malformed and non-numeric timing responses fail
- `git diff --check`

### DIFF
--- a/tinybird/README.md
+++ b/tinybird/README.md
@@ -81,8 +81,7 @@ Alternatively, create a more restrictive token in the Tinybird UI with `PIPES:RE
 benchmark from this directory:
 
 ```bash
-set -o pipefail
-TINYBIRD_HOST="$(jq -r '.host' .tinyb)" \
+TINYBIRD_HOST="$(jq -er '.host | strings | select(length > 0)' .tinyb)" \
 TINYBIRD_TOKEN='<copied token>' \
-./run.sh 2>&1 | tee log.txt
+./run.sh > >(tee log.txt) 2>&1
 ```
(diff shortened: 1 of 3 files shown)


tinybird/README.md
```diff
@@ -32,3 +32,4 @@ use the auto mode to make sure all the files are read.
 # Querying the data
 
 Once the data is inserted you can create the endpoints needed to run the benchmark using pipes. `run.sh` will iterate through each endpoint.
+Set `TINYBIRD_TOKEN` to a Tinybird token with read access before running the script.
```
rschu1ze: I created a Tinybird account but the UI is rather complex and this README covers it only rudimentary.

@fallintoplace  Since you seem familiar with Tinybird, maybe I could ask you to add
- the steps to install the Tinybird CLI in section "Creating an account" above,
- the steps to create and retrieve a "Tinybird Token" (l. 35),
- the steps to run the INSERT (l. 24),
- the exact invocation of `run.sh` (the PR message has an example)

rschu1ze (changes requested):

PR #954, opened 2026-06-22
Flag lukewarm cold runs and account for them in the Combined metric

toschmidt:
ClickBench's "cold run" is only meaningful when the data is actually read back from the storage device, i.e., the engine is restarted (or never persists between queries) **and** the OS page cache is dropped before the first run. Systems that instead run an unseen query against a still-warm engine, or managed/remote services whose caches can't be flushed at all, do a *lukewarm* run and are supposed to carry the `lukewarm-cold-run` tag.

A number of such systems were never tagged, which lets their warm "cold" numbers compete directly with genuinely cold systems, distorting the combined ranking. This PR flags them and reworks how the Combined metric treats lukewarm (and otherwise inapplicable) components.

### 1. `flag lukewarm cold runs`
Adds `lukewarm-cold-run` to systems that do not perform a true cold run:
- **Persistent local daemons never restarted between queries** (only the page cache is dropped, engine-internal caches survive): `impala`,  `firebolt`.
- **Managed/remote services** that can't be restarted or cache-flushed: clickhouse-cloud, motherduck, databricks, snowflake, redshift(+serverless), bigquery, athena(+partitioned), aurora-mysql, aurora-postgresql, alloydb, hologres, supabase, tablespace, tembo-olap, chyt, bytehouse, crunchy-bridge-for-analytics, timescale-cloud, tinybird, singlestore, hydra.

### 2. `reweight the combined metric for lukewarm and inapplicable metrics`
Unifies the per-metric exclusion rules in one `metricExcludes()` helper
(stateless→load, in-memory→cold/combined/load, lukewarm→cold, missing data
size→size) and reuses it everywhere:
- **Cold Run** ranking excludes lukewarm systems by default.
- The Combined per-query **cold baseline** excludes lukewarm/in-memory systems, so their warm "cold" numbers can't depress the baseline and inflate every true-cold system's cold ratio. `min` load time / data size likewise skip non-qualifying systems.
- A metric that doesn't apply to a system is **dropped, and the remaining weights renormalized** rather than fed a bogus ratio. Lukewarm systems keep a cold weight of 0 folded into hot (load 10% / size 10% / hot 80%).

### 3. `show individual hot/cold/load/storage ratios in the combined view`
Expands the Combined score cell to also list the four component ratios that feed it (`hot`, `cold`, `load`, `storage`), each as the relative `×N` ratio (`n/a` when a component doesn't apply).

FireShot Capture 060 - ClickBench — a Benchmark For Analytical DBMS -

### DIFF
--- a/alloydb/results/20231125/gcp.128GB_tuned.json
+++ b/alloydb/results/20231125/gcp.128GB_tuned.json
@@ -8,7 +8,7 @@
     "hardware": "cpu",
     "tuned": "no",
 
-    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp"],
+    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp", "lukewarm-cold-run"],
 
     "load_time": 0,
     "data_size": 9941379875,
--- a/alloydb/results/20231125/gcp.64GB.json
+++ b/alloydb/results/20231125/gcp.64GB.json
@@ -12,7 +12,8 @@
         "column-oriented",
         "PostgreSQL compatible",
         "managed",
-        "gcp"
+        "gcp",
+        "lukewarm-cold-run"
     ],
     "load_time": 1543,
     "data_size": 9941379875,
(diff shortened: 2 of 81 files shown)


toschmidt:
@rschu1ze can you take a look at the pull request?

PR #962, opened 2026-06-30
Read EventDate parquet column directly as DATE in Trino

raunaqmorarka:
Trino reads INT32-with-integer-annotation parquet columns as DATE since https://github.com/trinodb/trino/pull/25667, so the `date_add` view wrap on `EventDate` is no longer needed. Applied to all four Trino setups (`trino`, `trino-datalake`, `trino-partitioned`, `trino-datalake-partitioned`); validated against `hits.parquet` that results are identical.

### DIFF
--- a/trino-datalake-partitioned/create.sql
+++ b/trino-datalake-partitioned/create.sql
@@ -8,7 +8,7 @@ CREATE TABLE hive.clickbench.hits_raw (
     Title varchar,
     GoodEvent smallint,
     EventTime bigint,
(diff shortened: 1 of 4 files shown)


raunaqmorarka:
@alexey-milovidov @rschu1ze PTAL

rschu1ze (approved):

rschu1ze:
Verified to work. Not updating all Trino results because it is a minor change.

PR #969, opened 2026-07-01
hyper-parquet: Fix caching of Parquet files and add single-file variant

caetanosauer:
The hyper-parquet harness created a TEMP external table per ./query
connection. With the persistent-server model (issue #936) the driver
opens a fresh connection per try. This caused our plan cache to be wiped,
requiring re-sampling the data on every hot run as well. To properly
test a cached scenario, this commit changes our harness to create
persistent external tables and maintain an attached database throughout
the server's lifetime. The query iterations themselves still open a
fresh connection each time.

We also introduce the `single` variant for the parquet lane, now
reporting both `hyper-parquet-single` and `hyper-parquet-partitioned`.

As a drive-by change, we also drop ::bigint casts from ClickBench Q29,
since those are not needed anymore, matching the queries used by other
systems as well.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


caetanosauer:
Depends on #955 (persistent-server / issue #936). This PR is stacked on that branch — until #955 merges, the diff here also shows the persistent-server changes; it will collapse to just the parquet-caching work once #955 lands and this is rebased onto `main`.

PR #976, opened 2026-07-08
Add clickhouse-trash-schema benchmark

davidmenggx:
The same as ClickBench, but the `hits` table is split into a hundred tables with two columns each: `id` and one of the original columns, and `hits` becomes a view joining all of them by `id`.

See https://github.com/ClickHouse/ClickHouse/issues/109426


davidmenggx:
I wasn't able to run results yet, but I validated at small scale (schema creation, load and split, the `hits` view reassembling the original table exactly, all 43 queries)

rschu1ze:
> I wasn't able to run results yet, but I validated at small scale (schema creation, load and split, the hits view reassembling the original table exactly, all 43 queries)

I validated on a beefy machine c7a.metal-48xl and results were not great (lots of out-of-memory exceptions) :-( Anyways, I'll merge, we can use this as a starting point for optimizations. Thanks for working on this.

PR #985, opened 2026-07-13
Use octet_length instead of length in the DataFusion benchmark queries

### LINKED ISSUE #981: Use `octet_length` instead of `length` in the DataFusion benchmark queries
DataFusion's implementation of `length()` follows the SQL standard and is equivalent to `character_length()`, returning the number of Unicode characters.

However, the ClickBench queries currently use:

```sql
AVG(length(URL))
AVG(length(Referer))
```

For DataFusion, these expressions count characters rather than bytes. Computing character length requires scanning and decoding UTF-8, whereas `octet_length()` returns the number of bytes and is significantly cheaper.

The original ClickBench benchmark semantics appear to be based on byte length (e.g. ClickHouse's `length()` and DuckDB's `strlen()` return the number of bytes for UTF-8 strings), so using `octet_length()` in the DataFusion benchmark would better match those semantics while avoiding unnecessary overhead.

### Proposed change

In the DataFusion benchmark queries (Q27 and Q28), replace:

```sql
AVG(length(URL))
AVG(length(Referer))
```

with:

```sql
AVG(octet_length(URL))
AVG(octet_length(Referer))
```

### Benefits

* Preserves the intended byte-length semantics.
* Avoids unnecessary UTF-8 character counting in DataFusion.
* Makes DataFusion's ClickBench results more comparable with other engines.

For reference, the corresponding DataFusion issue is:
[https://github.com/apache/datafusion/issues/23086](https://github.com/apache/datafusion/issues/23086)

rschu1ze:
Resolves #981


rschu1ze:
That's a minor fix which likely doesn't change the benchmark results. I'll re-run the automation once there is a major DataFusion update PR.

alamb:
Thank you @rschu1ze and @kosiew -- much appreciated

PR #988, opened 2026-07-15
Support overriding the clickhouse binary with a PR build

alexey-milovidov:
## What

Adds a `clickhouse_pr` environment variable to the benchmark launcher. When set to a `ClickHouse/ClickHouse` pull-request number, the `clickhouse*` systems benchmark **that PR's CI release build** instead of the latest stable release. Every other system ignores it.

```bash
clickhouse_pr=81944 system=clickhouse ./run-benchmark.sh
```

## How

- `clickhouse_pr` is set on the operator side (`run-benchmark.sh`), threaded to the VM through `cloud-init.sh.in`, and consumed **only** by the `clickhouse*` install scripts.
- A new shared helper, `lib/download-clickhouse`, replaces the inline `curl https://clickhouse.com/ | sh` in every `clickhouse*/install`. With no `clickhouse_pr` set it keeps the historical stable-download behaviour; with it set it resolves and downloads the PR build.
- "The PR build" = the newest PR commit whose `Build (<arch>_release)` CI check finished **successfully** for the machine's architecture (`amd`/`arm`), then the raw self-extracting binary that build uploaded to S3:
  ```
  https://clickhouse-builds.s3.amazonaws.com/PRs/<PR>/<SHA>/build_<arch>_release/clickhouse
  ```
  Commits are walked newest-first, so a tip commit whose build is still running, failed, or was skipped falls back to the last commit that does have a finished build — i.e. "the last commit with a finished release build". Uses the GitHub API (honours an optional `GITHUB_TOKEN`); fails loudly with a clear message if no successful build exists.

## Filtering results

The chosen PR is written to the log as a `ClickHouse PR:` line and parsed by `sink.parser` into a **new `results.clickhouse_pr` column** so PR-build runs can be filtered apart from normal results (empty on a normal run). It is intentionally left out of the published `output` JSON, so the website payload is unchanged.

> [!IMPORTANT]
> `prepare-database.sql` is the schema *definition*; the live sink needs a manual migration to pick up the new column. SQL is in the PR discussion / provided to the operator.

## Scope

Touches the seven `clickhouse*` systems that download the universal binary: `clickhouse`, `clickhouse-trash-schema`, `clickhouse-web`, `clickhouse-parquet{,-partitioned}`, `clickhouse-datalake{,-partitioned}`. Leaves `clickhouse-tencent` (hardcoded PR build) and `clickhouse-cloud` (managed) alone.

## Testing

- Resolution dry-run against a real PR (#81944) for both `amd` and `arm` produced the correct, known-good S3 URL scheme.
- Ran the actual helper end-to-end with `wget` stubbed: native `arm` and forced-`x86_64`→`amd` both resolve correctly; no-`clickhouse_pr` falls back to the stable installer; missing-build case exits 1 with a clear message; candidate ordering is head-first, newest→oldest, deduped.
- `ClickHouse PR:` extraction verified: `81944` → `81944`, empty → `''`.
- All touched scripts pass `bash -n`; `cloud-init.sh` renders and validates with and without the PR set; the `sink.parser` SELECT (18 cols) aligns with the `sink.results` table.

### DIFF
--- a/clickhouse-datalake-partitioned/install
+++ b/clickhouse-datalake-partitioned/install
@@ -2,7 +2,7 @@
 set -e
 
 if [ ! -x ./clickhouse ]; then
-    curl https://clickhouse.com/ | sh
+    ../lib/download-clickhouse
 fi
 
 # Use a userspace page cache sized to ~80% of RAM for S3 object reads.
--- a/lib/download-clickhouse
+++ b/lib/download-clickhouse
@@ -24,7 +24,7 @@
 # back to the last commit that does have a finished build.
 set -euo pipefail
 
-# No PR requested: the historical default path.
+# No PR requested: the default path.
 if [ -z "${clickhouse_pr:-}" ]; then
     curl https://clickhouse.com/ | sh
     exit 0
--- a/lib/download-clickhouse
+++ b/lib/download-clickhouse
@@ -1,0 +1,94 @@
+#!/bin/bash
+# Put the `clickhouse` universal binary in the current directory.
+#
+# Default: download the latest stable build from clickhouse.com — exactly what
+# every clickhouse* install script did inline (`curl https://clickhouse.com/ |
+# sh`) before this helper existed.
+#
+# Override: when the `clickhouse_pr` environment variable names a
+# ClickHouse/ClickHouse pull-request number, fetch that PR's CI release build
+# instead of the stable one. This is how an operator benchmarks an unreleased
+# binary straight from a PR:
+#
+#     clickhouse_pr=81944 system=clickhouse ./run-benchmark.sh
+#
+# `clickhouse_pr` is set on the operator side, threaded to the VM through
+# cloud-init, and consumed ONLY here — so only the clickhouse* systems react to
+# it; every other system ignores it.
+#
+# "The binary from the PR" is resolved as: the newest commit in the PR whose
+# release-build CI job for THIS machine's architecture finished successfully,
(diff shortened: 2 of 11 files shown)


lib/download-clickhouse
```diff
@@ -0,0 +1,94 @@
...
+#
+# "The binary from the PR" is resolved as: the newest commit in the PR whose
+# release-build CI job for THIS machine's architecture finished successfully,
+# then the raw self-extracting binary that job uploaded to S3. Walking from the
+# newest commit backwards (rather than pinning the tip) means a tip commit whose
+# build is still running, failed, or was skipped (e.g. a docs-only commit) falls
+# back to the last commit that does have a finished build.
+set -euo pipefail
+
+# No PR requested: the historical default path.
```
alexey-milovidov:
```suggestion
# No PR requested: the default path.
```

clickhouse-datalake-partitioned/install
```diff
@@ -2,7 +2,7 @@
 set -e
 
 if [ ! -x ./clickhouse ]; then
-    curl https://clickhouse.com/ | sh
+    ../lib/download-clickhouse
```
alexey-milovidov, line 5: Let's embrace the simplicity of the default install.
So it should remain `curl https://clickhouse.com/ | sh` and be exposed here.

Add a non-standard setup only as a special case (a single if condition that invokes a script). Name the script to emphasize that it is non-standard.

Reason: no one should ever think that installing ClickHouse is not easy.

PR #993, opened 2026-07-20
Add GitHub Actions to launch benchmark machines on AWS

alexey-milovidov:
Automates launching ClickBench machines with GitHub Actions. The GitHub-hosted runner is used only to launch: it assumes a federated IAM role (OIDC) and runs the existing `run-benchmark.sh`, which starts a self-terminating EC2 machine. The machine clones the repository, runs `<system>/benchmark.sh` under cloud-init, sends its log to the results sink at play.clickhouse.com, and shuts down — exactly the current manual flow. `run-benchmark.sh` and `cloud-init.sh.in` are unchanged. Collecting results will be implemented separately.

### Workflows

| Workflow | Trigger | What it launches |
|----------|---------|------------------|
| `benchmark-daily.yml` | daily, 02:00 UTC | clickhouse, clickhouse-web, clickhouse-parquet, clickhouse-parquet-partitioned, clickhouse-datalake, clickhouse-datalake-partitioned, each on c6a.4xlarge, c6a.2xlarge, c6a.xlarge, c6a.large, t3a.small, c6a.metal, c8g.4xlarge, c8g.metal-48xl, c7a.metal-48xl, from main |
| `benchmark-manual.yml` | manual | any systems, machines, repository and branch |
| `benchmark-pr.yml` | pull requests | the systems whose directories the PR changes, from the PR's repository and branch, after approval |

All three share a composite action (`.github/actions/launch-benchmark`) that assumes the role and calls `run-benchmark.sh` in a loop, one launch per (machine, system) pair; a failed launch does not stop the rest and is reported at the end. The role ARN is set in the action, and the region defaults to us-east-1. The role's trust policy only admits workflow runs of this repository, so assuming it fails everywhere else, including forks.

### The PR workflow

- Uses `pull_request_target`: the workflow and launcher scripts always come from the trusted base branch, and OIDC works for PRs from forks. PR code never runs on the runner — only on the disposable benchmark machine.
- Detects systems to benchmark: top-level directories the PR touches (changes under `results/` and `*.md` files don't count) that contain `benchmark.sh` at the PR head — so newly added systems are included, and results-only PRs launch nothing. Verified against real PRs: #989 and #991 detect nothing (results-only), #987 detects the three Spice.ai systems.
- The launch job requires approval: it targets the `benchmark-approval` environment, which must be configured with required reviewers.
- A new push to the PR cancels runs that are still waiting for approval.

### Setup required (see `.github/workflows/README.md`)

[...]

PR #994, opened 2026-07-20
Record the pull request number in the results as clickbench_pr

alexey-milovidov:
Follow-up to #993: benchmark runs launched for a pull request are now marked, so the future collection step can tell them apart from the runs of main — without this, a PR run of an existing system would land in `sink.results` indistinguishable from the daily run and could be published as the official result.

- `cloud-init.sh.in` writes a `ClickBench PR: <number>` line into the log header (empty for runs of main).
- `sink.parser` extracts it into a new `clickbench_pr UInt32` column of `sink.results` (`prepare-database.sql`). The pattern is `(\d*)`, not `(\d+)`: the header line comes first in the log, and its empty match for runs of main stops the search before any spoofed `ClickBench PR: N` in untrusted benchmark output.
- `collect-results.sh` filters `clickbench_pr = 0`, so PR runs are never published.
- `repo` and `branch` are now empty by default: when `clickbench_pr` is set, `run-benchmark.sh` derives the PR's head repository and branch from the GitHub API; otherwise they fall back to `ClickHouse/ClickBench` and `main`. The PR workflow therefore passes only the PR number, and the manual workflow takes an optional `pr` input — any PR can be benchmarked by number.

Requires two statements on the sink database before merging (add the column, update the parser view); until the view is updated, PR runs would be recorded with `clickbench_pr = 0`.

### DIFF
--- a/.github/actions/launch-benchmark/action.yml
+++ b/.github/actions/launch-benchmark/action.yml
@@ -15,14 +15,21 @@ inputs:
     description: "EC2 instance types, space-separated"
     default: "c6a.4xlarge"
   repo:
-    description: "GitHub repository (owner/name) the machine will clone"
-    default: "ClickHouse/ClickBench"
+    description: "GitHub repository (owner/name) the machine will clone.
+      Empty: the PR's head repository if pr is given, else ClickHouse/ClickBench."
+    default: ""
   branch:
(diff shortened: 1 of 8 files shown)


alexey-milovidov (commented):
Ok.

PR #1009, opened 2026-07-20
Update Turso to v0.7.0, but it cannot finish the benchmark in reasonable time.

alexey-milovidov:
Bumps the pinned Turso CLI from `v0.1.2-pre.4` to the latest stable release, [v0.7.0](https://github.com/tursodatabase/turso/releases/tag/v0.7.0).

Only the version in `turso/install` changes. The v0.7.0 installer still creates `~/.turso/env` on a fresh machine, so the existing `source "$HOME/.turso/env"` lines in the scripts keep working unchanged.

Verified locally on 0.7.0:
- `./install` and `./check` pass (`tursodb --version` → Turso 0.7.0)
- `create.sql` DDL executes cleanly
- `.import --csv` still exists and a small import + query round trip works


alexey-milovidov:
Does not fit into the global timeout just a little.

alexey-milovidov:
With five minutes per query, it will likely need a few days to run.

PR #1128, opened 2026-07-28
Add CtrlB (Parquet, single)

asy62636:
Adds **CtrlB** in the Parquet, _single-file_ scenario, with results on c6a.4xlarge.

CtrlB is a query engine for observability data (logs, metrics, traces) that stores and reads data as Parquet. In this benchmark it queries the standard hits.parquet in place via CREATE EXTERNAL TABLE — there is no separate load step.

**How it runs:** CtrlB ships as a public Docker image (ghcr.io/ctrlb-hq/ctrlb-public, pinned in config.sh). The harness runs it as a long-lived HTTP daemon on port 5080 and POSTs each query to it. install pulls the image (and installs Docker if the VM doesn't have it); the run needs passwordless sudo (for drop_caches), same as the other locally-run systems.

**Timing / rules:** each query's reported time is the **client-side wall-clock of the full round-trip** — sending the query, the server processing it, and receiving the **entire** result body back — measured with curl %{time_total}. The full result is downloaded before timing stops; there is no compact/Null-format output suppression. Queries are the standard 43 (unmodified). concurrent_qps is left null — this config targets single-query latency, consistent with the other embedded/CLI Parquet entries.

Reproduce on a fresh c6a.4xlarge (passwordless sudo; Docker auto-installed if missing):

cd ctrlb
./benchmark.sh

Added the c6a.4xlarge result at ctrlb/results/20260728/c6a.4xlarge.json as requested (the fork is org-owned, so "Allow edits by maintainers" isn't available). Ready for review.


alexey-milovidov:
@asy62636, ok, I think we have enough results collected. Could you please copy-paste and commit them?

asy62636:
> @asy62636, ok, I think we have enough results collected. Could you please copy-paste and commit them?

@alexey-milovidov as requested, I have copy-pasted the results generated by the bot and committed it to the fork, as it should be visible to you. Apologies for being unable to grant "Allow edit by maintainers" on an org-owned repo. Could you kindly take a look once?

PR #1129, opened 2026-07-28
feat: add infino to clickbench

naman-modi:
Adds [infino](https://github.com/infino-ai/infino), a Rust engine that runs SQL, full-text (BM25), and vector search over one copy of your data on object storage. The data is spec-compliant Parquet on S3, Azure, GCS, or local disk, so anything that reads Parquet can read it. This entry runs the SQL path.

### Some Pointers:
- The harness is a single Rust binary that pulls only the published `infino` crate (0.1.10), with a committed lockfile and a plain release build, so a fresh VM reproduces it with nothing else to install. `queries.sql` is the same as the `datafusion` entry: the two type fixes datafusion does inline (day count to DATE, binary to UTF-8) are done at load time instead.
- infino stays up as a server between queries. `start` opens the table once and serves it over a unix socket, `query` is a thin client, and `BENCH_RESTARTABLE=yes` restarts it before each cold run. That meets the true-cold rule (cold first run, warm second and third), so there is no `lukewarm-cold-run` tag.
- Please Note: `load` includes infino compacting into its superfile format (256 MB segments). That is part of writing the data and is counted in `load_time`, not an optimize slipped in before the queries.

Ran the whole thing on a fresh c6a.4xlarge at 100M rows, built from crates.io: load, then the 43-query sweep with row counts checked. Hot sum 33.7s.


naman-modi:
Disclaimer: I work at Infino AI.

naman-modi:
Hey @alexey-milovidov / @rschu1ze, opened this to add infino. Would appreciate a review whenever you get a chance, happy to change anything. Thanks!

alexey-milovidov:
Build failed: https://pastila.nl/?0005853b/ad305de54c829af5dc2c24392b8dd3a4

naman-modi:
Fixed it @alexey-milovidov. `install` earlier assumed the runner VM had a C toolchain, but infino's build needs `cc`; it now apt-installs build-essential + pkg-config first. Would be great if you could re-trigger the run. Thanks!

naman-modi:
I have updated the results for `c6a.4xlarge` from [this](https://github.com/ClickHouse/ClickBench/pull/1129#issuecomment-5121001009) run in the PR.

naman-modi:
Hey @alexey-milovidov, it would be great if you could start the runs on other machines too? Thanks!

alexey-milovidov:
Thanks, waiting for results...

naman-modi:
Thanks Alexey!

PR #1132, opened 2026-07-28
Playground: bump "90+ databases" to "110+ databases"

alexey-milovidov:
Catalog is at 119 systems now; 110 are provisioned & queryable at the time of writing (72 snapshotted + 38 recently active). The "90+" number was from the original launch write-up.


alexey-milovidov:
Already merged.

PR #1159, opened 2026-07-30
Turso - it is ridiculously slow, can't believe that: revert the single-transaction import, cap per-query wall-clock

alexey-milovidov:
Follow-up to the failed turso runs on #1009. The bigger global timeout there was applied — cloud-init shows `+ timeout 72000 ./benchmark.sh` — but the run died at **6923 s**, nowhere near it, because the failure mode had changed.

## What the two logs say

**[2026-07-21](https://pastila.nl/?00170c10/0dcfdc73a284f71e075f528a0aba3d85)** (timeout 36000). Load succeeded in **6414 s**, four queries completed:

```
[321.512,308.438,308.561],   Q1 COUNT(*)
[344.419,331.820,331.802],   Q2
[352.985,340.337,340.340],   Q3
[340.837,328.204,327.412],   Q4 AVG(UserID)
Total time: 36020
```

Load + those four ≈ 10,600 s. The remaining **~7 hours** were Q5, `SELECT COUNT(DISTINCT UserID) FROM hits`, still running when the 10 h cap killed the run — discarding the four results it had.

**[2026-07-28](https://pastila.nl/?00170c10/5bf1e890010ce753f197c5561ae9fe42)** (timeout 72000). No `Load time:` line at all; `./load` never returned.

- Disk usage after 169,982,365,696 vs 19,471,605,760 before → **150.5 GB** written = `hits.csv` (74.8 GB, still present, so the `rm -f hits.csv` at the end of `load` never ran) + **~75.7 GB** of db+WAL. The 07-21 run's finished db was 68.9 GB, so the import had written essentially all the rows and went down at or near the `COMMIT`.
- No panic, no backtrace, no stderr despite `RUST_BACKTRACE=1` and `2>&1` into the pipe. That is a signal death. These machines run `earlyoom`; c6a.4xlarge has 32 GiB.

The change between the runs is #871, merged 2026-07-21 23:07 — *after* the 07-21 run, whose log still carries the unfiltered `Inserting batch of 1000 rows` lines. It wrapped `.import` in a single `BEGIN`/`COMMIT`, so tursodb 0.7.0 now accumulates the whole ~70 GB dataset in one uncommitted transaction.

## What this changes

**`turso/load`** — back to `.import`'s per-batch commits, keeping `PRAGMA synchronous = OFF` (the part of #871 that removes the per-batch fsync, without letting the write set pile up). The `wal_checkpoint(TRUNCATE)` goes too — there is no longer a 70 GB WAL to fold back in. Adds `set -o pipefail` and `|| true` on the progress filter: with per-batch commits every output line is a progress line, `grep -v` filters all of them and exits 1, and `set -e` would fail the load on that alone.

**`lib/benchmark-common.sh`** — new `BENCH_QUERY_TIMEOUT`, default `0` (no cap, existing code path unchanged for every other system). When set, a try past the cap is recorded `null` and the query's remaining tries are skipped, bounding a hopeless query at 1× the cap instead of 3×. Uses plain `timeout` (no `--foreground`) so the engine process dies with the wrapper instead of being orphaned. Also applied in the concurrent-QPS worker, which could otherwise block past its deadline and hang `wait` — a no-op today, since turso disables that test. The value is validated up front so a typo fails immediately rather than two hours in, after the load.

**`turso/benchmark.sh`** — `BENCH_QUERY_TIMEOUT=600`, ~1.7× the slowest query that has ever completed there (Q3, 353 s cold).

**`turso/timeout-seconds` + `run-benchmark.sh`** — the query cap alone is not enough: 43 queries × 3 tries at the ~350 s turso manages is over 12 h before the ~1.8 h load, so the 10 h default can only ever produce a dead machine. A system directory may now ship a `timeout-seconds` file; an explicit `timeout` in the environment still wins. Turso gets 20 h.

## Verification

Against a stub system: capped run gives `[0.203,0.203,0.202]` / `[null,null,null]` / `[0.303,0.303,0.304]` and continues past the timeout; the uncapped default still runs the hanging query three full times; no orphaned child processes after a cap fires; QPS worker terminates on a hanging query instead of hanging `wait`. Timeout resolution: turso → 72000, clickhouse → 36000, env override wins in both cases. Load pipeline: all-filtered-and-ok → 0, import failure → propagates, panic text still reaches the log.

Not verified on a real machine — that needs a ~2 h load plus the full sweep.

## Two things worth knowing

- The PR workflow checks out the **base** branch for the launcher scripts, so `turso/timeout-seconds` only takes effect once this is on main. Until then turso runs still need a manual `timeout` override.
- The expected outcome is a result row with **nulls** for the queries turso cannot do, not a clean sweep — Q5 and the other `COUNT(DISTINCT)` queries will almost certainly be null. That is the intent: a partial row beats a dead machine. It is not a "turso passes" result.


alexey-milovidov:
I need to give it 24 hours at least...

PR #1161, opened 2026-07-30
Add ClickHouse result for c7i.4xlarge

slice4e:
Adds a ClickHouse result for a new hardware configuration: AWS c7i.4xlarge (16 vCPU Intel Xeon Platinum 8488C, 30 GiB RAM), 500 GB gp2 root volume, Ubuntu 24.04.4, us-west-2d.

The environment matches cloud-init.sh.in: 16 GB swapfile, earlyoom enabled. The scripts in clickhouse/ were used unmodified — install, start, check, stop, load, query and data-size all as committed, driven by lib/benchmark-common.sh. ClickHouse version 26.8.1.440.

Cold runs are true cold runs: the server is stopped, the page cache dropped via drop_caches, and the server restarted before the first run of each query, so no lukewarm-cold-run tag applies.

metric	c7i.4xlarge	c6a.4xlarge (same date)
load_time	248 s	285 s
data_size	15,267,844,939	15,267,511,633
concurrent_qps	0.595	0.927
concurrent_error_ratio	0.707	0.578
Loaded 99,997,497 rows. ./validate-results.py . passes with no warnings for this file.

The run was operator-driven rather than launched through run-benchmark.sh, since that script provisions instances via the EC2 API.


alexey-milovidov:
What is the point?

slice4e:
> What is the point?

I do not see any Intel-based results (I work for Intel). Only ARM and AMD. It is a hardware configuration that is missing. I would like to contribute for completeness. If you think there is a better/different way - please let me know.

alexey-milovidov:
Let's add it here: https://benchmark.clickhouse.com/hardware/
The main ClickBench uses 9 different machines, and adding a new one will require running every system on it.

slice4e:
Thanks @alexey-milovidov - I see a number of different platforms evaluated there. I will see if I can contribute something meaningful there. ( still learning - thank you)

alexey-milovidov:
Continuation: #1218
Merged.

PR #1162, opened 2026-07-30
update umbra

toschmidt:
this is the same as #951 but with allow edits by maintainers enabled


toschmidt:
@rschu1ze @alexey-milovidov can you take a look at this, I created the pull request from a private repo so that you can push the results.

PR #1174, opened 2026-08-01
mongodb: fix the queries and the file the driver reads

alexey-milovidov:
Found while running the ClickBench MongoDB queries against the MongoDB compatibility layer of https://github.com/ClickHouse/ClickHouse/pull/68493. Four independent defects, all in `mongodb/`.

### `benchmark.sh` reads a file that does not exist

```sh
export BENCH_QUERIES_FILE="queries.txt"
```

There is no `mongodb/queries.txt`; the file is `queries.sql`, and there never was a `queries.txt` in this directory. `bench_run_query` reads `$BENCH_QUERIES_FILE`, so the driver got no queries at all.

### Q20 to Q23 lost their predicate

`queries.sql` was derived from `queries.js` with `EJSON.stringify`, which serializes a JavaScript regular expression literal as an empty document. So

```js
{ $match: { URL: /google/ } }
```

was committed as

```json
{"$match":{"URL":{}}}
```

which matches nothing — `{}` is an equality test against the empty document. The four affected queries are Q20, Q21, Q22 and Q23; they now carry the Extended JSON `$regularExpression` that `EJSON.parse` in `query` turns back into a regular expression. On 33.7M rows of `hits`, Q20 goes from an empty result to the 3979 rows the SQL returns.

### Q37 and Q28 did not implement their SQL

- **Q37** — the SQL is `... AND DontCountHits = 0 AND IsRefresh = 0 AND Title <> '' GROUP BY Title`, the pipeline filtered `URL: {$ne: ""}`. It is the `$match` of Q36 with only the `$group` changed.
- **Q28** — the SQL selects `k, AVG(length(Referer)), COUNT(*), MIN(Referer)`; the pipeline left out `MIN(Referer)`. Its regular expression was also `(?:www.)?` where the SQL has `(?:www\.)?`, so an unescaped `.` matched any character.

### The dates are not valid Extended JSON

`{"$date": "2013-07-01"}` is a bare calendar day. `mongosh` accepts it because it ends up in `new Date(...)`, but Extended JSON asks for a full instant and other drivers reject it outright — `pymongo`'s `json_util` raises `time data '2013-07' does not match format '%Y-%m-%dT%H:%M:%S'`, which makes the 7 date-filtered queries unusable outside `mongosh`. They are `{"$date": "2013-07-01T00:00:00Z"}` now, the same instant.

### `queries.js` is removed

`queries.sql` is the file the benchmark driver and the playground read. Keeping the same 43 pipelines a second time in `queries.js` and copying them across by hand is what dropped the regular expressions in the first place, so the copy is gone and `README.md` points at `queries.sql` instead.

### Verification

Every line of the new `queries.sql` was run against a MongoDB wire protocol endpoint over 33.7M rows of `hits` and compared with `clickhouse/queries.sql` on the same table. Q20, Q22, Q23, Q28 and Q37 now return exactly the values the SQL does, where before they returned nothing or the wrong ones. The queries that still differ are the ones that are not deterministic on this data — they `LIMIT` without a total order or cut through a run of ties, and their own SQL result changes between two identical runs — plus Q3, where `AVG(UserID)` overflows in the SQL and the pipeline's `$toDecimal` does not, and Q26, where the pipeline projects `EventTime` on purpose so that the following `$sort` can see it.

### The published results are now stale

Q20, Q21, Q22, Q23, Q28 and Q37 measured either nothing or the wrong work, so `mongodb/results/` should be re-run. I do not have a MongoDB installation to do that on the reference hardware.


alexey-milovidov:
We will merge without rerun. MongoDB is too slow.

PR #1307, opened 2026-08-10
feat: make infino runnable in the clickbench playground

naman-modi:
Makes the [infino](https://github.com/infino-ai/infino) entry run in the Playground too. It already exists in clickbench via #1129.

Some Pointers:

- Query now returns the result rows as a text table, not just a count. It is a bounded preview (first 1000 rows, capped at 256 KB) so a large result cannot blow up the VM. Timing is unchanged, it still wraps query_sql only and the table is formatted after the clock stops.
- The shared variables moved into a shared `config.sh` that load, start, and `benchmark.sh` all source, so the raw scripts get the same configs as benchmark.
- Bumped to the published infino crate `0.5.3` and pinned strong read consistency.
- Fits the 16 GB microVM with no swap: the full 100M load peaks around 11 GB.

### DIFF
--- a/infino/README.md
+++ b/infino/README.md
@@ -25,10 +25,12 @@ queries, so `tuned=no`.
 ./install        # installs rustup if absent, then cargo build --release
 ```
 
-The release build is `opt-level = 3`, `lto = "fat"`, `codegen-units = 1`
-(`bench/Cargo.toml`), a portable build with no `-C target-cpu`, matching the
-`datafusion` variant's recipe. Pinned to a published crate version
-(`infino = "0.1.10"`) so the result is reproducible from crates.io alone.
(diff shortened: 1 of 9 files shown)


naman-modi:
Hey @alexey-milovidov, can you please take a look at this PR? Thanks!

naman-modi:
Thanks for helping with this @alexey-milovidov! Should I commit the new results? They're within noise of the existing [20260730 entry](https://github.com/ClickHouse/ClickBench/blob/main/infino/results/20260730/c6a.4xlarge.json), which makes sense as this PR is just playground enablement, no perf change. I am fine either way.

alexey-milovidov:
Ok, not necessary to commit the new results. I'll merge this and refresh the Playground.

naman-modi:
Thanks @alexey-milovidov!

Issue #1266, opened 2026-08-07
How to update benchmark results? (for Hyper)


vogelsgesang:
We are about to release a new version of Hyper (probably later today) and would like to also update the ClickBench results.

I noticed that #993 put new automation in place.
Would you like us to still measure ourselves and create pull requests containing the new results?
Or should updates use the new `benchmark-manual.yml` Github flow? (We can't trigger flows ourselves, though, so we would need your support doing so)
Or should we pin the Hyper version in https://github.com/ClickHouse/ClickBench/blob/main/hyper/install, such that version bumps are done via explicit pull requests, which will trigger the `benchmark-pr.yml` Github flow?

StefanSteiner:
The new version of the Hyper API for August 2026 is now available through: pip install tableauhyperapi

rschu1ze:
> Would you like us to still measure ourselves and create pull requests containing the new results?

We now run all new and updated submissions unconditionally through the automation before merging them. You don't need to mess with the tags or the mechanics of the automation - we are happy to do that for you.

Your PR may include measurements (result xml files) or not, it's optional. Including results can be useful to ensure on your side that the scripts work.

> Or should we pin the Hyper version in https://github.com/ClickHouse/ClickBench/blob/main/hyper/install, such that version bumps are done via explicit pull requests, which will trigger the benchmark-pr.yml Github flow?

I'd say, if you bump the version in the install script and send that as a PR, it will be just fine. Thanks.

vogelsgesang:
#1308 - let's see if that works smoothly with the new Github flows 🙂

PR #1308, opened 2026-08-10
Pin Hyper version to 0.0.26225

vogelsgesang:
With the new infrastructure introduced in #993 benchmarks will be run via ClickBench-provided infrastructure instead of manually measured. Measurements are kicked off, e.g., by new Pull Requests.

This commit pins the Hyper version, such that version bumps in the script will automatically re-trigger the PR-automation remeasuring Hyper's performance numbers.

Also see #1266


caetanosauer:
@alexey-milovidov With the automation introduced in #993 , we expected this would run and merge automatically. Can you check why that was not the case? Looks like the GitHub action got stuck..

FYI @rschu1ze

PR #1357, opened 2026-08-13
Hyper: load native benchmark data from Parquet

caetanosauer:
Hyper: load native benchmark data from Parquet

Switch the native Hyper benchmark from the 81 GB CSV input to the 15 GB
single-file Parquet dataset. This reduces the measured load time from 327.0
seconds to 187.1 seconds (43%) while producing the same 99,997,497 rows and
the same logical results for all 43 ClickBench queries.


alexey-milovidov (approved):

vogelsgesang:
Are those results actually measured against this PR rebased on top of latest `main`? Do they also include  #1376 already? Or do we need to rebase the PR?

alexey-milovidov:
I remember I resolved a conflict and merged with master here: https://github.com/ClickHouse/ClickBench/pull/1357/commits/e235fccbbc6c0343b047c54807a9cebdb4f60201

Likely, they reflect the latest changes. If not, we can rerun.

Issue #934, opened 2026-05-19
Methodology: 3 hot iterations is insufficient for JVM-based engines


puzpuzpuz:
## Summary

The current methodology runs each query 3 times and reports `min(iter2, iter3)` as "hot" ([README "Caching"](https://github.com/ClickHouse/ClickBench#caching)). For engines that run on a JVM (Trino, Presto, Spark, QuestDB, Druid, Pinot, CrateDB, Doris FE, Greengage / WarehousePG, …), this window catches the JVM mid-warmup. The reported "hot" times are therefore systematically pessimistic compared to actual steady-state performance, by a measurable amount.

Proposing a discussion about increasing the default hot iteration count above 2.

## The mechanism

A JVM starts each query in the interpreter or low-tier compiled code, then progressively tiers up to fully-optimised native code as the JIT meets compilation thresholds. Methods can also be *deoptimised* when a speculation fails (e.g. a newly-loaded class invalidates an inlining assumption) and recompiled. All of this happens online, in microsecond-to-millisecond bursts, **inside query iterations**. Two consequences:

1. The first few iterations of any query are slower than the final steady-state by a factor that depends on how much code has tiered up. For engines that use per-worker thread pools, every worker crosses its own JIT thresholds independently - a 192-vCPU box has ~24x more workers than an 8-vCPU one, so the bigger box takes proportionally more iterations to fully warm.
2. JIT-compile and deopt events land at random points inside iterations. With only 2 hot iterations to choose from, the `min` definition gives lucky-or-unlucky outcomes: a single mid-iteration compile spike can push that iteration to 2-30x its steady-state time, and if it lands in iter 2 or 3, that is what ClickBench reports.

This is intrinsic to JVM runtime compilation, not a quirk of any single engine.

## Observed magnitude

Running a full 43-query suite for 10 hot iterations (instead of 2) on a 192-vCPU box exposes the gap between the 3-iter window and true steady state for one JVM-based engine in ClickBench:

| Hot iters | Total runs | Suite-sum (s) | Gap vs steady state | Queries within ±10% of steady state |
| --------- | ---------- | ------------- | ------------------- | ----------------------------------- |
| **2 (current ClickBench)** | **3** | **4.88** | **+48%** | **5/43** |
| 3         | 4          | 4.07          | +23%                | 19/43 |
| 4         | 5          | 3.61          | +9%                 | 24/43 |
| 5         | 6          | 3.48          | +5%                 | 34/43 |
[...]

PR #1358, opened 2026-08-13
Support variable query repetition counts

caetanosauer:
For the fastest systems in this benchmark, all queries run in under a second for the hot runs.
This leads to a high variability between different executions on the exact same system and hardware.
To alleviate this problem, this commit enhances the dashboard and associated metric scores to support
an arbitrary number of repetitions instead of the fixed 3. It aligns with the fact that the execution
scripts already support this through `BENCH_TRIES` (see `lib/benchmark-common.sh`).

Disclaimer: This leads to better results than currently published for our system (Hyper), which we
includ in this PR for the `c7a.metal-48xl` machine. We encourage
other systems to do the same so we can have more stable results overall.

This commit does not change the aggregate function, i.e., it still takes the min of the repetitions.
We might consider using median for more repetitions, but I left that out of scope here.

Note: this also addresses issue #934


caetanosauer:
@alexey-milovidov Curious to hear your thoughts on this one

alexey-milovidov:
Let's merge index.html for convenience, but don't change the results or benchmark rules for now.

There could be a lot of modifications to ClickBench, and all of them are easy to make right now:
- different number of repetitions;
- concurrent queries and measuring QPS;
- increasing dataset size;
- replacing Parquet files using the latest format optimizations;
- changing the set of queries or the number of queries;
- changing the dataset to another and adding more datasets.

But I want to avoid the "landscape problem".

PR #1375, opened 2026-08-14
Add Rayforce

alexey-milovidov:
Adds [Rayforce](https://github.com/RayforceDB/rayforce) — a zero-dependency
columnar analytics and graph engine in pure C (MIT), queried through **Rayfall**,
its Lisp-like language. `queries.sql` therefore holds 43 Rayfall expressions,
one per line, the way `jd` and `bqn` hold their own dialects.

(Not the same as the closed #940 — that was RayforceDB's internal 10M-row
harness, opened against this repo by mistake. This is a standard-harness entry
on the full dataset.)

## Two measured engine properties shaped the entry

**Text columns load as `SYM` (dictionary-encoded), not `STR`.** `STR` pool
offsets are `uint32_t`, so a column's pool is capped at 4 GiB, while at 100M
rows `URL`/`Title`/`Referer`/`OriginalURL` hold 5–11 GB of bytes each. Opening a
splayed table also validates every `STR` element, and that cost is superlinear:
0.4 s at 2M rows, **199 s at 9.6M rows** (99 s for the parted layout). All-`SYM`
opens in 17 s at 9.6M rows and stores the same subset in 4.4 GB instead of
7.5 GB.

**It runs as an IPC server, with `BENCH_RESTARTABLE=yes`.** The table open is
eager (validate every column file, load the dictionary), so a CLI-per-query
would pay it 129 times per run; the server pays it once per start and times
queries with `timeit` server-side. The restart is not optional either: with the
server left up, `drop_caches` cannot evict pages a live process still maps, and
a "cold" query measured 0.029 s versus 0.44 s after a real restart — keeping the
process alive would report every run as warm. `BENCH_CHECK_TIMEOUT` is raised to
1800 s because the readiness probe now waits for that open.

## Query notes

`HAVING` becomes an outer `select` over the grouped one; `LIMIT n OFFSET m` is
`take: [m n]`; `DATE_TRUNC('minute', EventTime)` is
`(xbar EventTime 60000000000)`; `CASE WHEN` is the row-wise `(if ...)`. Q29's
`REGEXP_REPLACE` is rebuilt from `str-find`/`substr`/`if` and reproduces the
regex exactly on this dataset, fallback rows included. Q25/Q27 project
`EventTime` alongside `SearchPhrase` because sorting runs after projection, and
Q35 projects its constant group key in an outer `select`.

## Validation

All 43 queries were diffed against `clickhouse-local` on a 2M-row subset of the
same CSV: 33 match exactly or modulo row order, and the other 10 differ only in
which of several equally-ranked rows a `LIMIT` over ties returns (the sort-key
multisets are identical). `install`, `load`, the stop/`drop_caches`/start/check
cold cycle, all 43 queries through `./query`, 10 concurrent clients, and the
error paths were exercised end to end.

## Upstream bugs found on the way

Neither affects this entry, both filed with minimal repros:

* RayforceDB/rayforce#404 — `select` with `take:` returns empty strings for
  pool-backed `STR` values (silently; `(select {from: t take: 1})` is enough).
  A further reason the text columns here are `SYM`.
* RayforceDB/rayforce#405 — ungrouped `(count (distinct col))` in a `select`
  projection returns the row count on `STR`/`SYM` and a type error on integers;
  hence Q5/Q6 use the whole-column reducer `(count (distinct (at hits 'c)))`.


alexey-milovidov:
It tends to randomly degrade up to being totally stuck:

```
[220.535000,3.190810,1.980040],
bench: ./check did not succeed within 1800s
bench: last ./check stderr was:
    error: io: connection refused: 127.0.0.1:5000
```

alexey-milovidov:
But as long as we have two configurations on which the system barely works, it's okay to merge.

alexey-milovidov:
But no - I don't think that the results are trustworthy. This needs a more detailed investigation.

alexey-milovidov:
I dug into the degradation. Short version: **the harness recognises errors correctly, the cold numbers are reproducible, and the hot numbers for two queries are not.** The cause is an upstream memory leak, not the entry — filed as [RayforceDB/rayforce#439](https://github.com/RayforceDB/rayforce/issues/439). I also found and fixed two real bugs in the entry's lifecycle scripts (7945c36a0).

## Errors are recognised correctly

I built v2.5.14 locally and exercised every failure path against the entry's own scripts:

| failure | what happens | recorded |
|---|---|---|
| server-side eval error | client aborts, exit 1, stderr `error: types` | `null` |
| Rayforce OOM | exit 1, stderr `error: oom`, server survives | `null` |
| server dies mid-query | `error: io: connection closed`, then `connection refused` on runs 2–3 | `[null,null,null]` |
| server never comes back | `bench_check_loop` fails, `set -e` aborts the run | no results file |

The logs match: Q12/Q29 are `[null,null,null]` after `connection closed`, Q33 is `[87.98,null,null]` after two `error: oom`. Nothing degraded gets silently recorded as a number. The `awk ms/1000` in `./query` *could* turn a non-numeric timing into `0.000000`, but it is unreachable — Rayforce aborts the client before anything is written.

## Root cause: a group-by hash table leaked per execution

Rayforce leaks memory proportional to the group count on every `select` with a **composite** `by:` key. Same query, six times, in one process, no IPC, no benchmark:

| repeats | `by: [a b]` (20M groups) | `by: a` (20M groups) | `by: b` (1000 groups) |
|--:|--:|--:|--:|
| 1 | 6.6 GB | 0.80 GB | 0.17 GB |
| 3 | 12.5 GB | 0.79 GB | 0.17 GB |
| 6 | 22.7 GB | 0.80 GB | 0.17 GB |

~3.2 GB leaked per execution, never returned; a single-key group-by over the *same* 20M groups is flat. It scales at ~160–230 bytes per group, so at ClickBench's ~100M groups that is roughly **16 GB per execution** on a 32 GB machine. `-m` does not contain it.

That is the whole story: the driver restarts the server before each *query* but not between the three *runs*, so run 1 is always clean and runs 2–3 execute in a process the previous run has poisoned. The queries that break are exactly the composite-key, high-cardinality ones — Q12, Q17, Q19, Q33.

## Which numbers survive

Comparing the two independent `c6a.4xlarge` runs already on this PR (74e151f0b vs 1db5f5ec3):

* **Cold: reproducible.** 38 of 43 queries agree within 1%; the only outlier is Q33 at 11.9%, the one that OOMs. Sums 228.0 s vs 238.0 s.
* **Hot: reproducible except two.** 41 of 43 agree within 2.5%, but **Q17 diverges 78.7%** (11.39 s vs 2.42 s) and **Q19 by 26.6%** (40.0 s vs 54.5 s). `min(t2,t3)` rescues Q17 in one run and not the other; Q19 has both runs degraded so nothing rescues it.

So it is not that the whole result set is noise — it is 2 unreliable hot numbers plus 3 error rows. Your call whether that clears the bar, but I would hold the results until #439 is fixed rather than publish two queries whose hot numbers are allocator state.

## Two bugs in the entry, fixed in 7945c36a0

**`./start` and `./stop` matched the server by process name.** `./query` and `./check` spawn clients that are *also* called `rayforce`, so `pgrep -x rayforce` cannot tell a live client from a live server. Deterministic repro: with a client alive and no server, `./start` exits 0, writes no `server.pid`, and `./check` fails. During the concurrent-QPS test ten workers keep a client alive most of the time, so the watchdog could silently fail to restart a crashed server; `pkill -x rayforce` in `./stop` also shot down the workers' in-flight queries on every poll. Both now use `pgrep -x -f "rayforce -p <port> server.rfl"`, which anchors to the whole command line and so avoids the `-f` hazard the original comment correctly warned about.

**`./check` called a server with no table "ready".** `server.rfl` has to tolerate a missing table because `./start` runs before `./load`, but it swallowed a *failed* open the same way: server up with `hits` bound to `0`, `./check`'s `(+ 1 1)` succeeds, and query 1 — `(count hits)` — returns `1` in two microseconds and is recorded as a valid timing while every other query errors out. Reproduced exactly. `./start` now sets `RAYFORCE_REQUIRE_TABLE=1` once `./hits` exists, and `server.rfl` exits rather than serve a table it could not open.

Verified locally: stays up pre-load with no table, refuses to start on a corrupt one, normal path unchanged, starts even with a client alive, leaves clients alone on `./stop`, still idempotent, and a watchdog can now revive a server killed under ten concurrent workers (error ratio 0.19 instead of 1.00).

## One driver-level thing I did not touch

`concurrent_qps` of 0.04–0.06 with `error ratio 1.000` is not a throughput measurement, and the leak alone does not explain it. Budgeting the run's wall clock — `Total time: 19003` s, minus load 3928 s, minus the 600 s window, minus 498 s of measured query time — leaves ~230–280 s per restart, i.e. **the eager table open takes about four minutes** on this dataset. That is why `BENCH_CHECK_TIMEOUT` is 1800 s here. But `bench_concurrent_qps`'s watchdog waits only a hardcoded 60 s after `./start` before looping round to `./stop`, so once the server dies during the window every restart gets killed a minute into a four-minute open and it can never come back. Making the watchdog honour `BENCH_CHECK_TIMEOUT` instead of 60 would fix it, but that changes behaviour for every system in the repo, so I left it out of this PR.

It is also worth stating plainly that with a four-minute open per query, roughly 3.3 of the run's 5.3 hours is table-opening that no reported number includes.

PR #1376, opened 2026-08-14
Use `octet_length` for Hyper, Umbra, CedarDB

vogelsgesang:
Switch Hyper, Umbra, and CedarDB to octet_length so Q27 and Q28 measure bytes, matching the intended ClickBench semantics while avoiding UTF-8 character scans when byte lengths are available from string offsets.

The queries for ClickHouse use `length(URL)` / `length(Referer)`. In ClickHouse, `length` is byte-oriented (https://clickhouse.com/docs/sql-reference/functions/string-functions#length), in contrast to `lengthUTF8`. Given ClickHouse itself uses a byte-length, the intent of ClickBench was probably to measure bytes, not UTF8. Hyper / Umbra / CedarDB were accidentally using UTF8 lengths, since their `length` function aligns with Postgres and nobody noticed the difference.

Also see #981, #985


vogelsgesang:
CC @toschmidt re the changes to Umbra
CC @lukasvogel re the changes to CedarDB

Please let know if you disagree with those changes. Happy to back them out of this PR, if you don't want them

alexey-milovidov (approved):

vogelsgesang:
just for reference (and so I can find it again later):

`c6a.4xlarge`, median of the two warm runs:
  
  | Database | Q27 | Q28 | Aggregate effect |
  |---|---:|---:|---:|
  | Hyper | 458 → 146 ms (**3.14×**) | 7.109 → 6.912 s (**2.8%**) | ~2.7% |
  | Umbra | 290 → 115 ms (**2.53×**) | 1.526 → 1.397 s (**8.5%**) | ~2.3% |
  | CedarDB | 275 → 108 ms (**2.54×**) | 3.198 → 3.097 s (**3.2%**) | ~2.2% |

PR #1396, opened 2026-08-15
Hyper Parquet: Disable sampling

vogelsgesang:
Disable creation of Parquet samples so the benchmark measures query execution rather than being over-shadowed by optimizer sampling


caetanosauer:
@rschu1ze Would you be able to have a look at this one? Much appreciated!

PR #1439, opened 2026-08-17
Make combined metric fairer and increase transparency

toschmidt:
1. The combined metric is difficult to interpret, so this PR displays the four individual values alongside it to make system strengths and weaknesses easier to identify.

2. Lukewarm-cold flagging: Several systems (especially cloud services) cannot be properly restarted, reporting lukewarm-cold times (new query, cached data) as true cold times (data loaded from storage). The PR marks several systems as `lukewarm-cold-runs`:  MotherDuck, Bigquery, ClickHouse Cloud, Redshift, Snowflake, ...
Note that existing ClickHouse Cloud results remain unchanged; the template has been updated so the next daily run will report results with the new flag.

3. Reweight the combined metric: if a system does not report a result for one of the four categories (hot, cold, load, storage), that category is ignored and the remaining score is rescaled.
If a system does not report storage/load the metric is rescaled: `exp( (0.6*log(hot) + 0.2*log(cold) + 0.1*log(storage)) / 0.9)` (if load is missing)
If a system does not report a cold time, we keep the 80%/20% split between query and load/storage and compute the combined metric as `exp(0.8*log(hot) + 0.1*log(load) + 0.1*log(storage))` (we could also remove this special case and just rescale like before)

image

### DIFF
--- a/alloydb/results/20231125/gcp.128GB_tuned.json
+++ b/alloydb/results/20231125/gcp.128GB_tuned.json
@@ -8,7 +8,7 @@
     "hardware": "cpu",
     "tuned": "no",
 
-    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp"],
+    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp", "lukewarm-cold-run"],
 
     "load_time": 0,
     "data_size": 9941379875,
--- a/alloydb/results/20231125/gcp.64GB.json
+++ b/alloydb/results/20231125/gcp.64GB.json
@@ -12,7 +12,8 @@
         "column-oriented",
         "PostgreSQL compatible",
         "managed",
-        "gcp"
+        "gcp",
+        "lukewarm-cold-run"
     ],
     "load_time": 1543,
     "data_size": 9941379875,
--- a/alloydb/results/20231125/gcp.64GB_tuned.json
+++ b/alloydb/results/20231125/gcp.64GB_tuned.json
@@ -8,7 +8,7 @@
     "tuned": "yes",
     "comment": "",
 
-    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp"],
+    "tags": ["C", "column-oriented", "PostgreSQL compatible", "managed", "gcp", "lukewarm-cold-run"],
 
     "load_time": 0,
     "data_size": 9941379875,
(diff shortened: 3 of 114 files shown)


toschmidt:
@alexey-milovidov what's your onpinion on this?

hall-alex:
> if a system does not report a result for one of the four categories (hot, cold, load, storage), that category is ignored and the remaining score is rescaled.

@toschmidt, this seems like a nice idea. What do you propose to do with cases where a system reports "unreasonable" times? E.g., load-time 2 secs. Would you ignore these as well?

toschmidt:
@hall-alex for systems with unreasonable load time or data size these results are excluded and are not factored in the combined metric. The metricExcludes function in index.html defines when a result is included. This function is also used to exclude systems from the load/size/cold views.

hall-alex:
> @hall-alex for systems with unreasonable load time or data size these results are excluded and are not factored in the combined metric. The metricExcludes function in index.html defines when a result is included. This function is also used to exclude systems from the load/size/cold views.

@toschmidt, ah, nice -- makes sense to me 👍

toschmidt:
@alexey-milovidov, @rschu1ze can you take a look at this?

hall-alex:
@toschmidt, it might be easier to get the changes reviewed if you split this into 3 (?) separate PRs. To my mind the "reweight the combined metric" seems like an easy and clear candidate to discuss. Why should systems that report almost 0 secs for one of the metrics get an "unfair" advantage in the combined metric? I like the simplicity and the idea of "not judging": one simply excludes the "unrealistic metric" from the combined score -- no need to explain / fix stuff.
Then again, I might be misinterpreting this -- perhaps it's by design that systems that, e.g., may hold the table readily available have a totally justified advantage over systems that first need to ingest 🤷‍♂️. Also, it is a hard call to judge what actually is "unrealistic" -- perhaps one shouldn't even go down that path? In any case that would probably be the first PR I would send out as an RFC.
@alexey-milovidov, what is your take on this?

(Side note: personally, I have found UX decisions like showing 4 metrics instead of 1 always hard to agree on.)

toschmidt:
@hall-alex yeah you are right. The UX change is debatable, I find my self switching between the different metrics quite often to see where systems perform well. However, 5 numbers instead of 1 might be confusing if you are not familiar with the benchmark. I will split the PR into 3 requests (first one is #1634).

I saw you updated the Firebolt submission, is the new version `lukewarm-cold` or is it a true `cold` now?

hall-alex:
We do not restart the system on each query, so we should be in the lukewarm bucket afaict ... I think that is the definition and I thought we were in that bucket.

PR #1461, opened 2026-08-18
Add StarRocks Parquet entries (single, partitioned) and Doris Parquet single variant

alexey-milovidov:
Adds three parquet-in-place entries:

## starrocks-parquet / starrocks-parquet-partitioned

Same local 1 FE + 1 BE cluster as the `starrocks` entry, but querying the parquet files in place through a view over the `FILES()` table function (`file://` protocol) instead of loading them into StarRocks storage. Two data-format quirks required fixups in the view:

- The files store `EventTime` as raw INT64 epoch seconds and `EventDate` as raw INT32 epoch days (no parquet logical types), so the view converts them to `DATETIME`/`DATE`. Without this, predicates like `EventDate >= '2013-07-14'` silently match **zero** rows.
- The partitioned files (unlike the single `hits.parquet`) lack the String annotation on their BYTE_ARRAY columns, so those infer as `VARBINARY` and string queries fail with `Error statistic type : varbinary`; the view casts them back to `VARCHAR` — the equivalent of duckdb-parquet's `binary_as_string=True`.

Validated locally on an arm64 box with StarRocks 4.0.2: all 43 queries pass on both variants, spot-checked results (including exact `COUNT(DISTINCT)` and the minute-truncation query) are identical to `clickhouse-local` over the same files, and the view survives the stop/start cold cycle the harness performs before every query.

## doris-parquet-single

Sibling of `doris-parquet` (which reads the 100 partitioned files) over the single `hits.parquet`, again via the `local()` TVF. One improvement over the sibling: `load` moves the file the harness already downloaded into the BE dir instead of downloading a second 14 GB copy. Not validated locally (Doris ships no arm64 tarball) — the PR benchmark run will cover it.


alexey-milovidov:
The first `doris-parquet-single` run failed at view creation with `Can not build local(): No available backend`. Root cause: the Doris start script (inherited from `doris-parquet`) waits for a non-empty `SHOW BACKENDS` column **by position**, which passes as soon as the backend row exists — so `./load` ran `create.sql` against a still-initializing BE. The `doris-parquet` sibling masks this race by re-downloading the whole dataset inside `./load` (~2.5 extra minutes), which this entry deliberately doesn't do.

Fixed in 697d7a53c: `./start` now waits for `Alive=true` (column located by header name, since positions shift between Doris versions) and fails loudly on timeout, and `./check` also requires an alive BE, so the harness' check loop gates the load and every per-query cold restart. The sibling `doris-parquet` carries the same latent race — worth porting the fix there if this run confirms it (left out of this PR to avoid re-benchmarking an unchanged entry).

alexey-milovidov:
Root cause of the `doris-parquet-single` failures found, and it affects every Doris entry in the repo: **cloud-init adds a 16 GB swapfile to all benchmark instances since f236e7ba2 (2026-05-09), and Doris's `start_be.sh` hard-refuses to launch while swap is active** (`Disable swap memory before starting be`). The launcher's message is invisible in the benchmark logs — it only surfaced after routing `start_be.sh` output into `./check`'s failure diagnostics. This is why neither `doris` nor `doris-parquet` has produced results since 2026-05-10.

Fixed here for `doris-parquet-single` with `swapoff -a` in `./install` (d93118b5d) — parquet-in-place doesn't need the swap headroom the swapfile exists for. The `doris` and `doris-parquet` entries need the same one-liner (plus ideally the Alive-gated `./check`/`./start` from this PR) in a follow-up.

PR #1462, opened 2026-08-18
keep the query column visible when scrolling the details table

toschmidt:
The detailed comparison table is wider than the screen, so scrolling right used to move the query checkbox and the query number out of view, leaving no way to tell which query a row belongs to.

### DIFF
--- a/index.html
+++ b/index.html
@@ -299,6 +299,29 @@
             padding: 0.1rem 0.5rem 0.1rem 0.5rem;
         }
 
+        /* The query checkbox and number stay put while the table is scrolled
+           horizontally, so it is always clear which query a row belongs to.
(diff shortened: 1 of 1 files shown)


toschmidt:
@alexey-milovidov can you take a look at this?

index.html
```diff
@@ -299,6 +299,29 @@
...
+           them the cells scrolling underneath shine through around it. */
+        #details .details-query {
+            position: sticky;
+            left: 0;
+            z-index: 1;
+            background-color: var(--background-color);
+            outline: 1px solid var(--background-color);
+        }
+
+        #details .details-query::before {
```
alexey-milovidov, line 315: Checked. Works great!

alexey-milovidov (approved):

PR #1463, opened 2026-08-18
detect errors in datafusion and fix datafusion vortex on c8g.metal-48xl

toschmidt:
Some queries in datafusion vortex partitioned failed due to the open file limit. The runner incorrectly reported the execution time of create table as query runtime leading to incorrect runtimes for some queries.

Errors are now detected correctly and the open file limit is increased before starting datafusion.


toschmidt:
@AdamGS do you agree with the change?

alexey-milovidov:
Thanks! I was also stumbling upon the numbers and found it suspicious.

alexey-milovidov:
Pushed 68021c4 onto this branch: the three historic `20260721` partitioned runs on the metal hosts still carry the mis-reported timings this PR prevents, so I replaced those non-measurements with `null` (Q20, Q21, Q27, Q28, Q33, Q34, plus Q22's cold try on c7a). Tries that actually completed are kept — c6a.metal Q28 run 3, c7a Q22 runs 2-3 and Q28 run 3.

`c6a.metal` and `c7a.metal-48xl` will now show gaps until they are re-run with the fix; `c8g.metal-48xl` is covered by the fresh `20260818` results in this PR.

AdamGS:
@toschmidt this makes sense to me, thank you for spotting it!

PR #1619, opened 2026-08-26
firebolt-parquet: fix the EventDate constants in Q36-Q42

alexey-milovidov:
## The bug

`firebolt-parquet/queries.sql` and `firebolt-parquet-partitioned/queries.sql` filter Q36–Q41 on `"EventDate" >= 16617 AND "EventDate" <= 16647`, and Q42 on `16630/16631`. Those are epoch-day values for July 2015 — exactly 730 days past the end of the dataset.

`hits.parquet` keeps `EventDate` as a raw day count, ranging 15888..15917 (2013-07-02 .. 2013-07-31), so **all seven queries match zero rows** and the timings recorded for them measure nothing:

```
SELECT COUNT(*) FROM hits
WHERE "CounterID" = 62 AND "EventDate" >= 16617 AND "EventDate" <= 16647   ->        0
SELECT COUNT(*) FROM hits
WHERE "CounterID" = 62 AND "EventDate" >= 15887 AND "EventDate" <= 15917   ->   738172
```

These two entries are the only ones in the repo using 16617/16630; every other entry that filters on the raw integer (`drill`, `octosql`, …) uses 15887/15917 and 15900/15901, which is what this PR switches to.

## Verification

Loaded `hits.parquet` into Firebolt Core both as a managed table and as a Parquet external table, and diffed all 43 query results. With the corrected constants, 34/43 match exactly — including Q36, Q37 and Q42 of the seven touched here. The remaining 9 (Q17, Q23, Q24, Q31, Q32, Q38–Q41) differ only in the order of tied rows under `LIMIT 10`, which is nondeterministic for every system in the benchmark.

## The results have to be re-run

The published results for both entries were produced with the broken queries, so their Q36–Q42 numbers (3–4 ms on `c6a.4xlarge`) are meaningless and optimistic. On this branch the same queries take 2–7x longer locally.

This is not confined to the two Firebolt entries: `Firebolt (Parquet) (c6a.4xlarge)` is currently the **global per-query baseline** for Q36, Q37, Q39 and Q42 at 3 ms, so on the untuned/hot view every other system is penalised by about 2.5% — Umbra (c8g.metal-48xl) 1.134 → 1.106, ClickHouse (c8g.metal-48xl) 2.337 → 2.279, and so on down the board.

The PR benchmark will re-run both systems on `c6a.4xlarge`. The existing results cover eight machines each, so a `machine:all` label is needed to replace all of them; alternatively the stale ones can be tagged `historical` until a full re-run lands.


alexey-milovidov:
The `machine:all` run failed on all 18 machines, but not because of the query fix — these two entries have been unable to produce results since May. From the `c6a.4xlarge` log:

```
Load time: 2.069
bench: data-size after load is '239646' (<5 GB)
bench: ClickBench's hits dataset doesn't fit in <5 GB on any
bench: system in the catalog; treating this as a partial load
```

Both entries copied `./data-size` from the ingesting `firebolt` entry, where it `du`s the engine's volume. These two ingest nothing — `create.sql` declares an external table over the Parquet files that `./load` leaves in `./data` — so `fb-volume` holds ~240 KB of catalog metadata and the partial-load guard from f26f504cf rejects the run before any query executes. The guard landed 2026-05-09, the entries moved to the shared driver 2026-05-15, which is why their newest results are still dated 2026-05-11 while `firebolt` has 2026-07-01 ones.

Pushed 3f4ec3b72:

- `data-size` now reports the source Parquet size, matching `clickhouse-parquet` / `duckdb-parquet` / `datafusion` / `chdb-parquet-partitioned` — and matching what the last published results for these entries contain (`14779976446` is exactly the size of `hits.parquet`).
- Dropped `BENCH_DURABLE=no`. That flag is for engines whose data lives in process memory, and it makes the driver re-run `./load` before every query and fold its wall-clock into the cold try. Here the Parquet files are on disk, the external table definition lives in `fb-volume`, and `BENCH_RESTARTABLE=no` means the container is never restarted — so the 43 extra loads would only have added ~2 s to every cold number. Every other external-table Parquet entry sets `BENCH_RESTARTABLE=no` alone.

Verified against a live `firebolt-core` container using this branch's scripts: `./load` succeeds, `./data-size` returns `14779976446` so the guard passes, and Q36–Q42 each return 10 rows.

Needs the `machine:all` benchmark re-run.

PR #1645, opened 2026-08-28
feat: update infino to v0.5.10

naman-modi:
Updates the [infino](https://github.com/infino-ai/infino) entry to crate `v0.5.10`, which decides how many superfiles a commit writes from the data committed rather than the host's core count, so the same load no longer writes far more files on a big host than a small one. load also passes a 1 second safety gap and that cost stays inside `load_time`.


naman-modi:
Hey @alexey-milovidov, can you please take a look at this PR when you get a chance, and add the machine:all tag if it looks fine? Thanks!

naman-modi:
Thanks for running machine:all, @alexey-milovidov! I have updated the results as per the run. 

All worked fine except `c8g.metal-48xl`. That one failed in setup, before infino ran: the Ubuntu ARM mirror (ports.ubuntu.com) returned 503 on the apt install of the build tools, so gcc never installed and the entry never built (63 fetch failures, aborted at 130s). Could you please re-run just `c8g.metal-48xl` when you get a chance? Thanks!

naman-modi:
Thanks @alexey-milovidov! I've added the `c8g.metal-48xl` result from the re-run, so all 9 machines are in now. This one looks good to merge from my side.

PR #1647, opened 2026-08-28
Rename the lukewarm-cold-run tag to no-cold

alexey-milovidov:
Follow-up to #1646. Same meaning, shorter name: the entry's first run of each query is not a true cold run, because the system is not restarted before it.

## Mechanical part

The literal `lukewarm-cold-run` → `no-cold` across:

- 29,852 result files
- `singlestore/template.json` (the only template that still carries it)
- the three scripts that stamp tags into new results — `clickhouse-cloud/collect-results.sh`, `databricks/benchmark.py`, `motherduck/benchmark.py`
- `README.md` section 2.b
- `data.generated.js`

The only occurrence of the old string left in the repo is the CHANGELOG entry recording the rename.

## Prose that needed more than a substitution

Two per-system READMEs referenced the tag in sentences that don't survive a blind rename:

- **`bemidb/README.md`** claimed the tag was set *"hence"* the cold run being genuinely cold — both self-contradictory and stale, since #1646 removed the tag from bemidb. Now says plainly that the driver restarts the daemon, so the cold run is a true cold run and no tag is set.
- **`pgrust/README.md`** said `(no `lukewarm-cold-run` tag)`, which renames into the double negative "no `no-cold` tag". Reworded to "so the `no-cold` tag does not apply".

## Notes

- `data.generated.js` is renamed in place rather than regenerated. A local `generate-results.sh` run would also pull in in-progress result directories that aren't ready to publish; the rename is a pure string substitution, so the website build step reproduces it exactly. Verified: 940 entries, 167 carrying `no-cold`, 0 carrying the old string.
- All 35,050 result files and templates re-parse as JSON, and `validate-results.py` reports the same 17 warnings / 1 error as before the change (all pre-existing).
- `README.md` section 2.b still explains the *concept* as "lukewarm cold runs"; only the tag name changed. Say the word and I'll rename the section too.


alexey-milovidov:
Just because it is simpler.

PR #1694, opened 2026-08-31
Add DolphinDB

alexey-milovidov:
Adds a `dolphindb/` entry. [DolphinDB](https://www.dolphindb.com/) is a closed-source columnar analytical and time-series DBMS written in C++, aimed at financial market data. It ships no SQL client binary — the download contains the server, a web notebook and the script interpreter, and everything else talks to it over the native binary protocol — so this entry drives it through the Python API, wrapped in `client.py`.

`create.sql` is real DolphinDB SQL DDL (`CREATE DATABASE` / `CREATE TABLE`), so the schema lives where ClickBench expects it. One DFS database, OLAP engine, `PARTITIONED BY EventDate` with a `VALUE` partition per day (30 partitions of ~3.3M rows, inside the 100MB–1GB range the vendor's partitioning guide asks for). No sort columns, no secondary indexes, no precomputation.

## Caveats

**1. The freely available license caps the server at 2 CPU cores and 8 GB of RAM per node, and enforces both.** `license()` reports `maxCoresPerNode = 2`, `maxMemoryPerNode = 8`, `bindCPU = true`, and the process pins itself at startup — `Cpus_allowed_list: 0-1` on a 96-core box. So on a c6a.4xlarge a run uses 2 of 16 vCPUs, and the fraction only shrinks on bigger machines. The server will not start at all without a license file (it exits silently). `./install` reads both caps out of `dolphindb.lic` rather than hardcoding them, so an Enterprise license dropped in its place is picked up — results obtained that way are a different configuration and would want a separate submission.

**2. No `results/` directory — please confirm you read the license the same way I do.** The download requires accepting the DolphinDB Software Evaluation License Agreement (`README.txt`: *"Before you install and use DolphinDB, you must read and agree to the included license agreements"*). §3 grants use *"internally within Your facilities solely for the purpose of evaluation"*, and §6 CONFIDENTIALITY adds:

> You may not use the Evaluation Software or any information relating to the Evaluation Software except for the purposes stated in Section 3 above.

No clause names benchmarks, but publishing measurements is not internal evaluation. So this takes the same route as `kdb`: reproduction scripts in full, `results/` gitignored, DolphinDB listed under *If The Results Cannot Be Published* in the top-level README, and `"dolphindb"` added to `_EXTERNAL` in the playground registry, since §5 forbids distributing "any portion of the Evaluation Software … to any third party" and serving it from a public VM would be exactly that. **If you read the agreement differently, all three of those are easy to reverse.**

**3. A completed 100M-row load has not been demonstrated.** The scripts run end to end, all 43 queries are verified, and a 3.1 GB slice (24 of the 605 segments) loads cleanly at ~17k rows/s — but the only machine available for the full run had its filesystem saturated by another tenant. With the shipped `dataSync=1` DolphinDB syncs on every segment commit, `sync()` waits on the whole filesystem's dirty pages rather than just its own, and single `syncAll` calls were taking 157 seconds. `dolphindb/README.md` has a *What has been verified* section stating this rather than implying the entry is proven at scale. Not expected on a dedicated VM with its own volume, but not shown there either.

**4. Loading needs `hits.csv`, and needs `atomic=false`.** Both for correctness:

- `loadTextEx` does not interpret ClickHouse's TSV backslash escapes. Nothing fails, the row count is right, and ~7300 `Title` / 520 `URL` / 375 `Referer` / 324 `SearchPhrase` values per million are silently different from what every other entry sees. The visible symptom is Q26 returning ten rows the reference doesn't contain, because search phrases beginning with `'` arrive as `\'`. `hits.csv`'s `"`-quoting loads faithfully, including the rows whose `URL` holds a raw newline inside the quoted field.
- `atomic=false` is documented as *required* when the file exceeds the cache engine capacity. Without it, an 81 GB load livelocks about six minutes in: resident memory pins to the 8 GB license ceiling and the log repeats `[TabletCache::flushContext] come across an exception : std::bad_alloc, and will retry later` once a second, indefinitely, and `SIGTERM` will not shift it.

## Query dialect

Seven differences, each a dialect issue rather than a change to what the query measures, and all written up in the entry README: `COUNT(DISTINCT x)` → `nunique(x)` (the former is rejected outright); no aggregates in `ORDER BY`; grouping expressions aliased in `GROUP BY` rather than `SELECT`; `LIMIT n OFFSET m` → `LIMIT m, n`, with Q39–Q43 wrapping the aggregation in a subquery because a DFS table refuses an offset combined with `GROUP BY`; `=` inside a function call being parsed as a keyword argument; `NOT` binding looser than `AND` (which silently returned 354 rows where the reference returns 70 until Q23 was parenthesised); and the function-name mapping (`strlen`, `minuteOfHour`, `bar(t, 60)`, `regexReplace`). Only Q35 could not keep its shape, since `GROUP BY 1` is rejected and grouping by `URL` alone is equivalent.

All 43 were diffed against `clickhouse-local` running `clickhouse/queries.sql` over the same 1% sample: 30 match row for row, 4 return the same rows in a different order among equal sort keys, and the remaining 9 differ only where the queries permit it — Q18 has no `ORDER BY`, Q32/Q33 group by `WatchID, ClientIP` where every group has count 1, and Q10's `AVG` differs in the fifth significant digit.

## Also

`./stop` waits for the listening socket to be released rather than only for the process to leave, and skips zombies; `./start` reports an immediate exit on stderr. Both were added after a thread stuck in `sync_inodes_sb` survived `SIGKILL`, kept port 8848 bound, and turned the next start into `Failed to bind the socket on port 8848` — which `bench_start` discards, leaving only an unexplained 300 s readiness timeout. The driver does 43+ stop/start cycles, so one lingering socket would fail a whole run.

`./start` also raises the open-files soft limit to 102400, which the vendor's machine-preparation tutorial asks for.


alexey-milovidov:
Ok. We can't publish results, but at least it can be in the repository as a template.

PR #1703, opened 2026-08-31
Add Regolith

vertexclique:
Adds [Regolith](https://sourcenetwork.github.io/regolith/) which is ACID KV-engine for edge systems.

Benchmark submission is identical to how you did the SlateDB's submission. @alexey-milovidov 

Looking forward to see the results 🚀


vertexclique:
I think this needs an approval again idk why @alexey-milovidov 😄 probably for other machines

alexey-milovidov:
It looks at the new commits and asks for approval, but the approval is already done on the previous commit, and the results will arrive.

PR #1724, opened 2026-09-01
Firebolt: keep the three entries in the no-cold bucket

hall-alex:
Firebolt does not handle the start / stop cycle from #1666 gracefully as of Sept 2026, so those runs do not lead to meaningful/interesting measurements. This restores `BENCH_RESTARTABLE=no` and the `no-cold` tag on `firebolt`, `firebolt-parquet` and `firebolt-parquet-partitioned`.

Note that we **continue to flush caches** as all other `BENCH_RESTARTABLE=no` systems, see [here](https://github.com/ClickHouse/ClickBench/blob/main/lib/benchmark-common.sh#L266).

Also adds a `c8g.metal-48xl` result measured in that configuration.

The result files #1666 produced were true cold runs and keep no tag; new runs will carry it again.

Additionally, switches the `firebolt` entry from `engine:dev` to `engine:dev_mv1`, a build we are verifying and re-tagging by hand, so that any mainline regressions cannot walk into the benchmark unnoticed.


hall-alex:
@alexey-milovidov, can you please take a look?

alexey-milovidov:
> does not handle the start / stop cycle

How exactly?

hall-alex:
@alexey-milovidov, thanks a lot for the fast response!

> How exactly?

Essentially, we are developing for an "always up" type of usage. E.g., certain per table metadata takes time to load -- we don't have (and do not plan on) optimizing for super-fast truly cold queries. Some of this noise creeps into the 2 warm runs as well. I think the "no-cold" bucket simply makes more sense for firebolt as of today.

If this gives the default view with the combined metric an "unfair picture" of Firebolt (and other no-cold systems), you could consider [this PR ](https://github.com/ClickHouse/ClickBench/pull/1634). I hope I'm not misunderstanding the intention of that PR, but I **think** this is about not counting the cold runs or the load-times for systems that don't actually measure "true cold runs" or "true load-times". Please correct me if I'm wrong.

alexey-milovidov:
The intent was to restart every system. Even in-memory data frame systems are restarted, they just have to load the dataset from scratch for every query and count it as the cold query time. It is okay that systems have to load metadata at start. However, there is a loophole - the startup time is not counted, iirc, so lazy loading is penalized unfairly. I will give more thought to this.

hall-alex:
> It is okay that systems have to load metadata at start.

@alexey-milovidov, I guess we could look into this on our end, but I'd really prefer for now to just put Firebolt back into `no-cold` (i.e., the main setup we're aiming for) and possibly have the combined metric 1) exclude these cold-runs for all no-cold systems and 2) also exclude the load-times for systems which do not measure a full ingest -- as the mentioned PR suggests.

The current PR + the one with the combined-metric fix would result in the following default view:
image

PR #1726, opened 2026-09-01
Add OceanBase (column store and row store)

alexey-milovidov:
OceanBase is the shared-nothing, Paxos-replicated, multi-tenant DBMS Alibaba started in 2010 and open-sourced under Mulan PSL v2 in 2021: a MySQL-compatible SQL layer — `SELECT VERSION()` answers `5.7.25-OceanBase_CE-v5.0.1.0` — over an LSM-tree storage engine. Since 4.3 that engine can store a table by column instead of by row, and the two are different enough here to justify two entries: `oceanbase` is the column store, which is what the vendor's own OLAP parameter template selects, and `oceanbase-row` is the same setup with `WITH COLUMN GROUP (all columns)`.

It installs natively; no container is needed. OceanBase publishes only el7/el8 RPMs, but the observer links nothing outside glibc and a `libaio` that ships in the `oceanbase-ce-libs` package, so `rpm2cpio | cpio` into `/opt/oceanbase` runs unmodified on Ubuntu — on x86_64 and aarch64 alike, so the `c8g` machines are covered too. Nothing is installed system-wide.

### What was surprising

- **The whole cluster setup has to live in `./install`.** A freshly started observer serves no tenant at all — it answers `Tenant not in this server` until `ALTER SYSTEM BOOTSTRAP` has created the internal `sys` tenant, and user data then needs a tenant of its own, which needs a resource pool, which needs a unit config. All of that has to finish before the driver's first `./check`.
- **`./load` has to force a minor freeze of the *meta* tenants, or the entry produces no result at all.** The direct load leaves ~100 MB of redo log on the meta tenant that every user tenant carries, and that stream's `base_lsn` stays at zero, so every `./start` replays the whole thing. How much replay a tenant can buffer is bounded by its memtable, and a meta tenant's memtable is about 4% of the resource unit's memory — 410 MB on a 9 GB unit. The backlog does not fit, replay stalls with `CLOG pending size in task queue exceeds limit`, the observer never reaches `start success`, and `./check` times out on all 43 queries. Measured on the unit whose meta tenant gets exactly that 410 MB: without the freeze the server had not finished starting after ten minutes, twice; with it, 27 seconds. `TENANT = all` covers the sys and user tenants — the meta tenants need `all_meta`.
- **Load parallelism has to be capped by memory, not just by cores.** The vendor's rule is "up to the tenant's core count", but each direct-load worker holds a sort area, a macroblock writer and a 7 MB coroutine stack; at 90 workers against a 9 GB tenant the load reaches ~10 GB of data and dies with `ERROR 4013: No memory or reach tenant memory limit`, rolling everything back. `install` caps it at one worker per 512 MB of tenant memory, which on every machine this benchmark runs on is a no-op.
- **`secure_file_priv` can only be set over a Unix socket**, never over TCP, so `install` uses `-S /opt/oceanbase/run/sql.sock` for that one statement.
- **`du` is useless for the data size.** The observer preallocates its data file and its redo log pool at startup, so `du` reports the reservation — 159 GiB against 21.1 GiB of real content on the full dataset. `./data-size` reads `DATA_DISK_IN_USE + LOG_DISK_IN_USE` from `GV$OB_SERVERS` instead: 11.9 GiB of columns plus 9.2 GiB of redo log.
- **`VARCHAR`, not `TEXT`, and the widths are not free.** `TEXT` is a LOB type in OceanBase and the benchmark's hottest columns are `URL`, `Title` and `Referer`. `VARCHAR` needs a declared width, and OceanBase caps a row at 1.5 MB of *declared* width, which 28 `VARCHAR(65535)` columns exceed fivefold. The widths are at least 4× the longest value the dataset actually holds, measured in the loaded table.

### Configuration

`install` sizes the instance the way `obd`, the vendor's deployer, does — `memory_limit` at 80% of RAM, `system_memory` from obd's step function, `cpu_count` at `nproc - 2` — plus the `olap` parameter and system-variable template that ships inside the RPM itself (`etc/default_parameter.json`, `etc/default_system_variable.json`) and that obd and OCP offer as a dropdown. Two of those entries change results and not just speed: `utf8mb4_bin` makes `LIKE` and `ORDER BY` byte-exact, which is what the reference ClickHouse results do, and `parallel_degree_policy = AUTO` is what lets a query use more than one thread without a hint in the SQL. `install` refuses to run below 8 GB of RAM, the vendor's documented minimum.

### Query results

`queries.sql` differs from `clickhouse/queries.sql` on two lines out of 43: Q29's `REGEXP_REPLACE` backreference has to be `'$1'` (OceanBase follows MySQL 8, where `'\1'` is the literal character `1`, and left alone the query collapses every row into one group), and Q43 groups by minute, which the `mysql` entry's `'%H:00:00'` gets wrong.

Every query was compared against `clickhouse-local` on the same rows. 33 of 43 agree exactly; eight are `LIMIT` cutting through tied sort keys (verified — the sort-key multisets match, and in Q24 the ten `WatchID`s are the same ten); Q4 differs because *ClickHouse* overflows `AVG(UserID)` in an `Int64` while OceanBase returns the exact decimal mean; and Q6 differs by 2 in 107907 because `utf8mb4_bin` is a `PAD SPACE` collation and two phrase pairs differ only in a trailing space.

### Verification

Run end to end on the full dataset: `./load` gets exactly 99,997,497 rows in, the `APPEND` hint's online statistics land with them (`num_rows = 99997497`), all 43 queries return a result, `./stop` → `./start` → `./check` → query works, and the driver's own `bench_run_query` and `bench_concurrent_qps` were run against the loaded table. The rowstore entry is validated at 1% scale.

The machine that ran it is a shared 96-core box, not a benchmark VM, so no timing here means anything — and at 96 cores against a 9 GB tenant, a tenth of the memory per core that any machine in this benchmark has, `AUTO` parallelism picks a degree the tenant cannot afford for Q19, Q32 and Q33. Pinning the degree to 13 — what a 16-core machine would choose — runs all three in 22.6 s, 74.6 s and 40.9 s.

No results yet; those need runs on the benchmark's own EC2 machines.


alexey-milovidov:
> What was surprising

This is laughable. How not to build a database :rofl:

alexey-milovidov:
```
[2026-09-01 19:25:19.226034] INFO  [LIB] pidfile_test (utility.cpp:1113) [2439][observer][T0][Y0-0000000000000000-0-0] [lt=0] fid file doesn't exist(pidfile=run/observer.pid)
ERROR 4015 (HY000) at line 1: System error
```

This is pathetic.

`INFO  [LIB]` - double whitespace
`[Y0-0000000000000000-0-0]` - what's that shit?
`[lt=0]` - what's those?
`fid` - what's "fid"?
`exist(pidfile=run/observer.pid)` - missing space before `(`
`ERROR 4015 (HY000) at line 1: System error` - system error :rofl:

alexey-milovidov:
It must be a very serious enterprise database. Judging by error messages.

PR #1887, opened 2026-09-07
Add Ravel

pmoust:
[Ravel](https://github.com/NOFireAI/ravel) is an object-storage-native telemetry database queried over SQL (and other DSLs). 
Every durable byte lives in S3-compatible object storage, no WAL.

Two results on c6a.4xlarge, both true cold: the server is restarted and the page cache dropped before the first run of every query, 3 runs per query. The local disk holds only the downloaded Parquet file and a read cache.

The stock entry runs the server with no performance flags. 42 of 43 queries complete; q33 exceeds the per-query memory pool the server derives on a 30 GB machine and is recorded as null rather than run with a raised limit. 

The tuned entry passes four flags through RAVEL_TUNED_ARGS in ./start: latency-first fetch policy, fetch concurrency 256, a 12 GiB per-query pool, and a 26.3 GB read cache (80% of this machine's memory) so the dataset stays resident between the three runs of a query. All 43 queries complete. It is marked tuned and sits out of competition; the README states the flags and the trade they accept, about 5.3x the object-store GET requests of the stock entry for the shorter cold wall-clock.
We're working to improve the price-perf of that mode, as well as providing adaptive modes and dynamic query-cost budgets. But it's safe to say that there always going to be a "give me best latency, no matter what, in that architecture" mode. This is that mode right now.

Load time is the shipped recipe at defaults and was measured separately from the query runs, since the load is not idempotent against an already loaded bucket. It reliably measures in the same exact w/ <+-0.5% deviation with fresh buckets.

What's different with Ravel in the context of ClickBench is that it requires an S3 bucket to ingest to (and read from). 
Is that something you can facilitate @alexey-milovidov? 
Happy to follow any route you feel appropriate.


alexey-milovidov:
This makes sense. We can start with a local instance of Minio or SeaweedFS, so the setup will be comparable with other systems. Then we can extend the benchmark to support systems on real S3 (but currently it isn't supported).

pmoust:
Done: the entry now runs against a local single-node MinIO, in the second commit on this branch.

`./install` downloads a pinned MinIO release (checksum-verified), generates a credential pair into a file readable by the current user only, starts MinIO on loopback with its data directory on the machine's disk, creates the bucket and qualifies it. Nothing is required from the operator; `./benchmark.sh` runs as is. A non-loopback `RAVEL_S3_ENDPOINT` skips the MinIO setup so the same scripts can run against another S3-compatible store later.

Both results are re-measured on that setup, same c6a.4xlarge, same true-cold protocol (server restarted and page cache dropped before the first run of every query). Load time 1,203 s.

- Stock: no flags. Cold 1,701 s, hot 184 s; q33 exceeds the derived per-query memory pool and is recorded as null, as before.
- Tuned: the same four flags as before plus the read cache's local-disk tier (`RAVEL_CACHE_DIR`). Cold 727 s, hot 82 s, all 43 queries. The tier sits on the same volume as MinIO and survives the restart before each cold run, so a query's cold run benefits from objects earlier queries touched; the README says so, since it is part of what those figures measure.

For context on the cold numbers: Ravel keeps no data on local disk, so every cold run reads from MinIO, which reads from the 500 GB gp2 volume at about 256 MB/s. The stock policy reads the whole 11.2 GB corpus per query, which is about 44 s per query at that rate whatever the query computes. MinIO also serves a ranged GET by reading every 1 MiB block the range touches, which is why the tuned policy's byte savings show up on the wire more than on disk without the tier. The earlier real-S3 figures stay in the README as a reference table only, not as a results file.

pmoust:
Pinned the entry to Ravel v0.15.0, which is now published, in the commit above.
The earlier numbers were taken on code that had not shipped yet; `./install`
downloaded the previous release, so the benchmark run measured that. The
released binaries are what it installs now, and the collector's next run
measures them.

Two smaller things in the same commit. The 0.15.0 server records query text on
its audit trail tokenized under a key and refuses to start unkeyed without one,
so `./install` generates a key into the same private credentials file as the
MinIO keys. And the tuned result file is not re-added: the collector removes
hand-added results for a system it benchmarks, so the README now presents those
flags as a configuration rather than a published result, with the reference
table saying which build each row came from.

On our own machine the released binaries measure stock 1,702 s cold and 184 s
hot, tuned with the disk tier 728 s and 83 s, load 1,224 s, all within noise of
the previous run.

alexey-milovidov:
@pmoust, it can't install:
> curl: (22) The requested URL returned error: 410

PR #1904, opened 2026-09-08
Hyper: Pin new version 0.0.26558 and add Web variant

caetanosauer:
This commit updates Hyper to a newer version and introduces a new "Web" variant. Inspired by the same variant in ClickHouse, this reads the Hyper database file on demand at query time directly from a public S3 bucket.

We tried to follow the blueprint of `ClickHouse (Web)` here, but please let us know if some step is incorrect. Furthermore, we rely on `curl` to fetch the object data size from S3 directly.

For our standard `hyper` lane, we also switch the storage backend to `psync`, which gives a more representative and stable load time for the particular EBS volume used here. Unlike the previous backend, it does not rely on the kernel to flush pages at unpredictable times.


caetanosauer:
@rschu1ze @alexey-milovidov sorry for the extra ping here, but I would be very thankful if we could merge this soon. We want to showcase this Web functionality internally very soon, and ClickBench performance is as always a big reference for us.

Let me know if there's anything I can do to help. Much appreciated in advance!

rschu1ze (approved):

PR #2066, opened 2026-09-17
umbra: update to the 26.09 image

toschmidt:


alexey-milovidov:
Thanks! I tried very hard to make it run on https://benchmark.clickhouse.com/playground/, but nothing helped. Hope it will run after this change.

toschmidt:
> Thanks! I tried very hard to make it run on https://benchmark.clickhouse.com/playground/, but nothing helped. Hope it will run after this change.

I was able to run get it running locally with the playground setup but I am not sure what the problem on the official release is. So not sure it will work this time.

PR #2082, opened 2026-09-18
Add MonetDB result for c6a.2xlarge (2026-09-18)

KazukiKandaKK:
Existing MonetDB results were from 2022-07-01 on c6a.4xlarge only. This adds a fresh run on c6a.2xlarge (8 vCPU / 16 GiB, gp3 220GB) using MonetDB 11.55.7 via the official APT repository, following the standard `benchmark.sh` flow (43 queries, 3 runs each, plus the concurrent QPS follow-up).

- Load time: 1693.08s (99,997,497 rows)
- Data size: 48,831,609,113 bytes
- Concurrent QPS: 0.038, error ratio: 0.000

Validated with `validate-results.py` (exit 0, no errors/warnings for this file).


alexey-milovidov:
@KazukiKandaKK 

```
monetdb: cannot connect: control socket does not exist
bench: ./check did not succeed within 300s
```

KazukiKandaKK:
@alexey-milovidov Thanks for running this. I hit the same `control socket does not exist` / `./check` timeout on a fresh Ubuntu 24.04 box with the stock `monetdb/` scripts, so I dug into it before resubmitting.

On MonetDB 11.55.7 (current APT package), `monetdbd` refuses to run as root. From the official Dec2025-SP3 (11.55.7) release notes (https://www.monetdb.org/release-notes/):

> It is no longer allowed to run mserver5 or monetdbd as root. There are important security concerns when these servers are run as root.

`install` does `sudo monetdbd create /var/lib/monetdb 2>/dev/null || true` and `sudo monetdbd start /var/lib/monetdb 2>/dev/null || true` (same in `start:25`):
https://github.com/ClickHouse/ClickBench/blob/3aebf663ae768fe3e961fcae0aed337e5a4318e1/monetdb/install#L18-L19
https://github.com/ClickHouse/ClickBench/blob/d9ef173d031a6b2c2a08a431f3c9b7b61db12ac2/monetdb/start#L25

so that root-refusal failure is silently swallowed. No daemon comes up, `monetdb create test` has no control socket, and `./check` (just `mclient -h 127.0.0.1 -d test`) waits 300s for a server that never started.

Also, the packaged systemd unit's dbfarm is `/var/monetdb5/dbfarm`, not `/var/lib/monetdb` (`install` already notes `/var/lib/monetdb` is the `monetdb` user's home). Per MonetDB's own docs:
- https://www.monetdb.org/documentation-Dec2025/admin-guide/server-setup-and-configuration/ — "The default location of the dbfarm for a daemon managed by systemd is `/var/monetdb5/dbfarm`."
- https://www.monetdb.org/easy-setup/ubuntu-debian/ — "the databases will be in `/var/monetdb5/dbfarm`"

`start`/`stop`/`data-size` still hardcode the home dir path. Not the main failure, but it's the same scripts.

The numbers in this PR are from starting the daemon via systemd, so `monetdbd` runs as the `monetdb` user instead of root. That's what actually made the run complete. I only have the CI log links, not a shell on the runner, so I can't confirm that's what happened there. It does reproduce locally, and it matches the root restriction and dbfarm path documented above for this MonetDB version.

Happy to send a fix for `install`/`start`/`stop`/`data-size` that runs `monetdbd` as non-root (systemd, or a dedicated user if systemd isn't there) and points at the real dbfarm. Can put it on this PR or a separate one, whichever you prefer.

alexey-milovidov:
🕵 @KazukiKandaKK Your diagnosis is right, and I've pushed the fix to this branch so the run can be redone here. Reproduced it on a MonetDB 11.55.7 (Dec2025-SP3) install from the same APT repo:

```
# monetdbd create /var/monetdb5/dbfarm
fatal: running as root is not allowed
```

Three commits:

**[`da4a379`](https://github.com/ClickHouse/ClickBench/commit/da4a3792e) — run the servers as the `monetdb` user, on the packaged dbfarm.**
Everything daemon-side now goes through `sudo -u monetdb`, and the dbfarm moves to `/var/monetdb5/dbfarm` as you suggested. Two things beyond what you listed:

- `./stop` needed it too — `monetdbd stop` is refused as root just the same. So the cold cycle's stop was a *silent* no-op even when the daemon did come up, and the "cold" run was served by a process that never went away. That is why this entry still carries the `no-cold` tag. With the fix `./stop` really does take down `monetdbd` and `mserver5`, `./check` starts failing, and the cold run is a true cold run — so results produced from now on should not have the tag (`template.json` doesn't set it, so the automation's files won't).
- `install` now disables the packaged `monetdbd.service` before starting the daemon itself. The unit has `Restart=on-failure`, so a systemd-managed daemon comes straight back after the `SIGTERM` from `./stop`, and the cold cycle would quietly stop being cold again.

Also: `install` now fails with a diagnostic when the daemon never comes up instead of falling through to a bare `./check` timeout, `load` makes the whole path down to `hits.tsv` traversable (`mserver5` is no longer root, so it no longer bypasses directory permissions), and `query.expect` is gone — nothing has called it since the entry moved to driving `mclient` directly.

**[`46844fb`](https://github.com/ClickHouse/ClickBench/commit/46844fb35) — parse mclient's minute and hour clock formats.**
This one affects your result file. `mclient -t clock` switches unit as the duration grows (`clients/mapiclient/mclient.c`):

```
clk: %ld.%03d ms
clk: %ld.%03d sec
clk: %ld:%02d min
clk: %ld:%02d h
```

`query` fed the value to awk's numeric conversion, which reads `2:05` as `2`. So every query slower than a minute was recorded as its *minute count* in seconds:

| mclient printed | old `./query` | fixed `./query` |
|---|---|---|
| `clk: 923.456 ms` | 0.923 | 0.923 |
| `clk: 12.345 sec` | 12.345 | 12.345 |
| `clk: 1:14 min` | **1.000** | 74.000 |
| `clk: 1:30 h` | **1.000** | 5400.000 |

In `monetdb/results/20260918/c6a.2xlarge.json` that is Q24 (`[2.0, 4.95, 3.778]` — the cold run was 2:0x, not 2 s, which is also why it reads as faster than its own warm runs) and Q33 (`[1.0, 1.0, 1.0]` — the `GROUP BY WatchID, ClientIP` query, 22.8 s on the 2022 c6a.4xlarge entry, so 1:xx on half the cores is exactly what you'd expect). Verified end to end against a query timed at `clk: 1:14 min`, which the fixed script reports as `74.000`.

**[`d8f6559`](https://github.com/ClickHouse/ClickBench/commit/d8f655935) — give the post-load restart room to replay the WAL.**
`COPY INTO` writes the ingested data to the write-ahead log — 200k rows of hits.tsv left a 142 MB `sql_logs/sql/log.1` next to only 86 MB of BATs — and nothing truncates it until the database shuts down cleanly. Since this entry never restarted before, that cost was never paid; now the first cold cycle pays it. `BENCH_CHECK_TIMEOUT` goes to 3600 s (only the first restart needs it), and monetdbd's `exittimeout` to 7200 s so a slow clean shutdown isn't `SIGKILL`ed into skipping the checkpoint and replaying the log on *every* cold cycle.

**Testing.** All of the above was exercised on an amd64 Ubuntu 24.04 install of MonetDB 11.55.7: `./install` from a clean slate and re-run for idempotency, `./start` / `./check` / `./load` / `./data-size`, a full `./stop` → `./start` round trip with the data still there, all 43 queries through `./query`, and `benchmark.sh` end to end against a 200k-row slice of hits.tsv (cold cycle per query included).

I've launched the benchmark on `c6a.2xlarge` and `c6a.4xlarge`. When it produces results the automation commits them here and removes the hand-added `monetdb/results/20260918/c6a.2xlarge.json`, so the two timer-corrupted rows go away with it.

PR #2083, opened 2026-09-18
Add ObsessionDB

marcmacmac:
### Add ObsessionDB

ObsessionDB is a ClickHouse-based database with separated compute and storage: data on
object storage, a metadata layer and a distributed NVMe cache shared across the
compute nodes. The table engine is our own implementation, built against the open-source
`SharedMergeTree` API.

**22 results**: seven service sizes (8, 16, 32, 64, 128, 256, 336 GiB) at 1, 2 and 3 nodes, plus
one single-node dev instance.

#### Method

The entry is deliberately a copy of `clickhouse-cloud`, because the two systems are operated the
same way and we did not want any ambiguity in the comparison:

| file | difference from `clickhouse-cloud` |
|---|---|
| `run.sh` | none — byte-identical |
| `queries.sql` | none — byte-identical |
| `create.sql` | none — byte-identical |
| `benchmark.sh` | header comment (our console URL), and `urlCluster('cluster', ...)` instead of `urlCluster('default', ...)` — `cluster` is the cluster name on an ObsessionDB service |

We picked up #2046, the `urlCluster` source and the `EXPECTED_ROWS` guard. The row-count guard is a good addition and we kept it.


alexey-milovidov:
I've logged in to ObsessionDB, but can't find a button to create services with these sizes. Could you please share screenshots or detailed steps?

marcmacmac:
Hey @alexey-milovidov, thanks for checking. It's available for our enterprise customers only. I added four clusters with two nodes each for the 8, 16, 32 and 64GB tiers to your account.
Let me know if you want to verify the full sweep on all seven tiers with one, two and three nodes.

Here's how you can access it

On the top left select your service
image

In the left navigation select "Connect" and there you'll find your connection details.

Let me know if you have any questions with it.

alexey-milovidov:
I'll check it. But the results should be reproducible by a user "from the street" to publish.

marcmacmac:
That's fair. I appreciate your effort anyhow. Our public self-service is around the corner. So I'd ping you once everyone can verify it "from the street".

PR #2085, opened 2026-09-18
Add intent-gizmosql (GizmoSQL fork with DuckDB engine changes)

ZENGXH:
Adds `intent-gizmosql`: a released open-source fork of GizmoSQL v1.38.0 — the same server as the existing `gizmosql` entry — whose embedded DuckDB v1.5.5 carries a small series of engine changes.

**Source and release**
- https://github.com/bddppqs/intent-gizmosql, tag `v1.38.0-clickbench.1`: the release this entry installs (portable Linux amd64 build, glibc 2.28 baseline, `SHA256SUMS` attached).
- https://github.com/bddppqs/intent-duckdb, tag `v1.5.5-clickbench.1`: the DuckDB changes, described in that repository's `CLICKBENCH-FORK.md` (batched RE2 character-class runs, CountZeros builtins with selection-fed HyperLogLog, an evictable decoded-dictionary cache with admission back-off, a cross-request memo of literal `regexp_replace` domains over dictionary vectors). They do not change query results, the storage format or the SQL surface; the server runs any SQL DuckDB accepts.
- Both repositories keep their upstream licenses (Apache-2.0 / MIT).

**Relation to the `gizmosql` entry**
- Every script is the upstream `gizmosql` entry's, byte for byte, except `install` (downloads the pinned release zip and verifies its SHA-256 instead of running the vendor's network installer; no loader shim on Ubuntu 24.04) and `query` (fails a run only on the client's own diagnostics: the upstream check also matched result rows, and a URL containing the word "error" inside Q39's tied-count window produced a null timing on one of our test machines).
- Default server configuration, the upstream schema (`create.sql`) and load path; no index, pre-aggregation or query-specific setting.

**Disclosure and labels**
- The engine changes were developed and evaluated against the ClickBench queries; the entry is labelled `tuned: yes` as requested in review. If you prefer a different name or tag, please relabel it.

**Reproduction**
- The committed result comes from `benchmark.sh` run unattended by cloud-init on a fresh c6a.4xlarge (Ubuntu 24.04), no manual steps (load 121 s; the full run takes about 25 minutes).
- The release is amd64 only, so ARM machines will not produce results for this entry.


intent-gizmosql/template.json
```diff
@@ -0,0 +1,11 @@
+{
+  "system": "intent-gizmosql",
+  "proprietary": "no",
+  "hardware": "cpu",
+  "tuned": "no",
```
alexey-milovidov: Should be "yes".

ZENGXH: Done in 984227d: `tuned` is now "yes" (template.json and the committed result file). The same commit makes `query` fail a run only on the client's own diagnostics: the upstream script also matched result rows, and a URL containing the word "error" inside Q39's tied-count window turned a valid run into a null timing on one of our test machines.

ZENGXH:
Committed the result of your run as `intent-gizmosql/results/20260919/c6a.4xlarge.json` in e26a9de and removed our manually added 20260918 file.

ZENGXH:
Committed all six result files under `intent-gizmosql/results/20260919/` in 1545c05 (the c6a.4xlarge file is the later run from the machine:all batch). The two c8g runs fail as noted in the description: the release is amd64 only.

ZENGXH:
Committed the t3a.small result as `intent-gizmosql/results/20260919/t3a.small.json` in 007bcde. That completes every amd64 machine type; the two c8g runs stay empty (amd64-only release).

